PR #33, opened 2015-01-14
Custom port configuration

nickkorn:
Allow the user to specify the server port when creating a Remotery global instance. Allows a user to run multiple instances of a program on the same computer and view it in Remotery or deal with port conflicts.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -3898,7 +3898,7 @@ static rmtError Remotery_ThreadMain(Thread* thread)
 }
 
 
-static rmtError Remotery_Constructor(Remotery* rmt)
+static rmtError Remotery_Constructor(Remotery* rmt, int port)
 {
     rmtError error;
 
@@ -3918,7 +3918,7 @@ static rmtError Remotery_Constructor(Remotery* rmt)
         return error;
 
     // Create the server
-    New_1(Server, rmt->server, 0x4597);
+    New_1(Server, rmt->server, port);
     if (error != RMT_ERROR_NONE)
         return error;
 
@@ -4078,14 +4078,14 @@ static void Remotery_DestroyThreadSamplers(Remotery* rmt)
 }
 
 
-rmtError _rmt_CreateGlobalInstance(Remotery** remotery)
+rmtError _rmt_CreateGlobalInstance( Remotery** remotery, int port )
 {
-    rmtError error;
+	rmtError error;
 
-    // Creating the Remotery instance also records it as the global instance
-    assert(remotery != NULL);
-    New_0(Remotery, *remotery);
-    return error;
+	// Creating the Remotery instance also records it as the global instance
+	assert( remotery != NULL );
+	New_1( Remotery, *remotery, port );
+	return error;
 }
 
 
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -50,6 +50,8 @@
 
 #ifdef RMT_ENABLED
 
+// Global settings
+static rmtSettings g_Settings;
 
 
 /*
@@ -3791,7 +3793,8 @@ static rmtError Remotery_ConsumeMessageQueue(Remotery* rmt)
         return RMT_ERROR_NONE;
 
     // Loop reading the max number of messages for this update
-    while (nb_messages_sent++ < MAX_NB_MESSAGES_PER_UPDATE)
+	const rmtU32 maxNbMessagesPerUpdate = g_Settings.maxNbMessagesPerUpdate;
+	while( nb_messages_sent++ < maxNbMessagesPerUpdate )
     {
         rmtError error = RMT_ERROR_NONE;
         Message* message = MessageQueue_PeekNextMessage(rmt->mq_to_rmt_thread);
@@ -3888,7 +3891,7 @@ static rmtError Remotery_ThreadMain(Thread* thread)
         // This loop will exit with unrelease samples.
         //
 
-        msSleep(MS_SLEEP_BETWEEN_SERVER_UPDATES);
+        msSleep(g_Settings.msSleepBetweenServerUpdates);
     }
 
     // Release all samples to their allocators as a consequence of [NOTE-A]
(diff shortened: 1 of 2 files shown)


dwilliamson:
Thanks for the pull request but can you do this another way?

You're already thinking along the same lines as I am: I don't want early adopters to have to think about any initialisation parameters when initially integrating (hence a separate initialisation function being a good idea).

However, I want to do this through a global "settings object" which is initialised with default values on startup. Something like this:

```cpp
rmtSettings* settings = rmt_Settings();
settings->port = 0x1234;
```

Stuff I want to put in this settings object includes:

* Connection port.
* Hostname to limit connections to.
* Memory use (see https://github.com/Celtoys/Remotery/blob/master/lib/Remotery.c#L148).
* Control over thread behaviour (e.g. how often the server sleeps).
* Function pointers for malloc/free to allow custom memory management.
* Filename to log recorded data to (a planned feature of Remotery).

For now all we need is the port number.

To keep ```#ifdef``` requirements out of the client code I think the settings object should be defined even if ```RMT_ENABLED``` is not.

nickkorn:
I can probably set up a good number of those things in the next day or two.

Just some quick clarification. What would the call to rmt_Settings() do in your mind? I'm guessing the first but just wanted to check before I start modifying anything.

static rmtSettings* g_RemoterySettings = NULL;
rmtSettings* rmt_Settings()
{
  if( g_RemoterySettings == NULL)
    New_0(rmtSettings, g_RemoterySettings);

  return g_RemoterySettings;
}

or

static rmtSettings g_RemoterySettings;
rmtSettings* rmt_Settings() { return g_RemoterySettings; }

dwilliamson:
I don't have a strong preference for either right now. Some comments on each:

* New
 * Can use NULL to check whether settings need to be initialised to their default value on first call.
 * Doesn't have a means through which the allocated memory can be released unless a new function is added. I don't like really like this. Binding it to the global remotery object isn't possible without making the init API more complicated (that and I want people to be able to restart the server without losing their settings object).
* Static
 * Predictable memory use.
 * Requires a flag somewhere to check and see if the settings object has been default-initialised.
 * Triggers C/C++ zero-initialisation for the entire struct (which may be a good thing) and embeds itself in the executable image.

I guess the second idea wins?

nickkorn:
Quick thing regarding the static as I'm starting to look into it:
In Remotery_Constructor I will need to validate each parameter to ensure it has been initialized to a default value or you're doing if value is 0 checks everywhere. Having a flag to determine if it is initialized or not doesn't serve much purpose unless we are going to have the user tell us they have initialized all of the values. 

In the New case this is handled by the new function to default all values and then the user modifies the ones required. If the user hasn't specified anything then the Remotery_Constructor just creates the settings and they are constructed with default values.

In the case of Remotery being disabled:
Probably want to write RMT_OPTIONALs for modifying the settings parameters.
Utilizing the "New" version would require a smaller memory footprint, although maybe the compiler just optimizes out the "Static" version if it just exists and isn't put to use?

nickkorn:
I made some code changes to reflect what was discussed here using the static method. Seems to work well with Remotery enabled or disabled. 

I was going to look into adding allocate/free functions to the settings struct as well. The one issue I ran into with doing that is BeginNew is called before I can fill in the default values in Remotery_Construct. One way to handle that situation would be to fill in the default values for the Settings in _rmt_CreateGlobalInstance. Not sure if that is how you planned on handling those functions so I left it out for the time being.

dwilliamson:
I was thinking that keeping the settings object completely disconnected from the remotery object would be the simplest and cleanest solution for now, that doesn't change the API. So:

```cpp
struct rmtSettings
{
   rmtU16 port;
};

static rmtSettings g_Settings;
static rmtBool g_SettingsInitialised = RMT_FALSE;

rmtSettings* rmt_Settings()
{
   // Default-initialise on first call
   if (g_SettingsInitialised == RMT_FALSE)
   {
      g_Settings.port = 0x4597;
      g_SettingsInitialised = RMT_TRUE;
   }

   return &g_Settings;
}

static rmtError Remotery_Constructor(Remotery* rmt)
{
   // Call right at the start to initialise settings in-case the client hasn't
   rmt_Settings();
}
```

In the case of malloc/free function pointers this then doesn't become an issue because both the C and C++ standards guarantee zero-initialisation of global statics at start-up. You only pay the cost of the size of the static variable in the executable image's data segment. This leads to a static malloc call like:

```cpp
static void* Malloc(rmtU32 size)
{
   if (g_Settings.malloc != NULL)
      return g_Settings.malloc(size);
   return malloc(size);
}
```

It's a little slower than just having ```g_Settings.malloc``` point directly to CRT ```malloc``` but then there are no malloc calls on the fast path in remotery.

nickkorn:
That all sounds good to me. I'll make those changes in a bit.

dwilliamson:
Actually, what was I thinking.. it can be a little simpler than that - just call ```rmt_Settings()``` from rmt_CreateGlobalInstance``` ! From what I can see there are no ```malloc``` calls before that point so default initialisation can point to the CRT version and there's no need to check for NULL anymore.

nickkorn:
That route works for me. Just wasn't sure if you had a reason to keep the contents of that function minimal.

dwilliamson:
That's almost perfect... could you change a couple of things:

* Convert the tabs in your changes to spaces (4 per tab).
* ```Remotery.h``` doesn't require any dependencies before you include it. The new ```malloc``` pointer changes that. Is there any way to point to malloc without ```size_t``` defined? If not, the fallback would be to include ```stddef.h```

nickkorn:
Not quite sure there is a great way to determine size_t that will work across compilers without a bunch of preprocessor tests per-compiler.

Another fallback consideration would expect a 32-bit or 64-bit integer input and then write a wrapper function that calls malloc on the source side.

dwilliamson:
Of course! I like that second idea as that could be extended a little for people with memory allocators that need to store context:

```cpp
typedef void* (*rmtMallocPtr)(void* mm_context, rmtU32 size);
typedef void (*rmtFreePtr)(void* mm_context, void* ptr);

struct rmtSettings
{
   rmtMallocPtr malloc;
   rmtFreePtr free;
   void* mm_context;
};

// In remotery.c
static void* Malloc(rmtU32 size)
{
   return g_Settings.malloc(g_Settings.mm_context, size);
}
```

So now users can pass pointers to functions that use a specific pool or whatever. For the CRT ```malloc`` the ```mm_context``` pointer can be left as NULL and the user would write:

```cpp
void CRTMalloc(void* mm_context, rmtU32 size)
{
   return malloc((size_t)size);
}

rmtSettings* settings = rmt_Settings();
settings->malloc = CRTMalloc;
```

I'm a bit slow at the moment, working late :)

nickkorn:
No worries, this has been my primary task for the day so I've been a little more focused on it. Hopefully this is saving you work not causing you more by having to verify everything.

dwilliamson:
Yes, thanks, this helps quite a bit.

PR #34, opened 2015-01-21
Bugfix linux gcc

Nodrev:
Various compilation fixes for linux platform.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -2746,7 +2746,7 @@ static Message* MessageQueue_AllocMessage(MessageQueue* queue, rmtU32 payload_si
         rmtU32 s = queue->size;
         rmtU32 r = queue->read_pos;
         rmtU32 w = queue->write_pos;
-        if ((int)(w - r) > s - write_size)
+        if ((int)(w - r) > ((int)(s - write_size)))
             return NULL;
 
         // Point to the newly allocated space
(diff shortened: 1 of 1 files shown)


dwilliamson:
Thanks for the fixes, I'm just a little unsure of 928b643. Can you give me more info on what that achieves?

Nodrev:
As stated in the commit comment, this avoid the following warning:
warning: variable ‘rmt’ set but not used [-Wunused-but-set-variable]
when RMT_ENABLED is not setted, by forcing a dummy use of "rmt" variable.
This is not really a bugfix, but I like when code do not produce any warning :)
So feel free to take this commit or not :)

dwilliamson:
Right, but what's the context of this; where is your ```rmt``` variable stored? I can see this being a problem:
```
void TestFunction()
{
   Remotery* rmt;
}
```
as it's a local variable that goes unused. But does using a global or class member variable for the ```Remotery*``` pointer trigger the same warning too?

Also, how does ```RMT_UNUSED_PARAMETER``` suppress that warning given that it's a pre-processor macro that resolves to no generated code?

It's all confusing me a little :)

dwilliamson:
If you could create another pull request for  0bafe45, 410caba and bc90494 I'll merge them immediately, while I figure out that last commit.

Nodrev:
Branch modified, I'll create another pull request for the last commit.

dwilliamson:
Thanks.

Do you know if the email you build your commits with matches the email you have associated with your Github account?

If you look at your contributions in the commit list, we can't click on your account name to find out more about you https://github.com/Celtoys/Remotery/commits/master . All of your commits won't link to your account (for example 8e263a43689365fbffc36cd27badaed32b72043b )

There should be a blue question mark with a circle next to your name in the last link that tells you how you can fix this.

PR #35, opened 2015-02-01
Linux x86 timer and thread validity

barrettcolin:


dwilliamson:
Thanks for these.

Could you look at your change on the main branch: a7d03685d770ec0a23832f52c23a82f09a35e9f2

Next to your name is a blue question mark that says Github can not link the commit to your Github account. Is the email you submit that commit with the same as your Github one? It means currently you're not linked as one of the contributors to the repo.

barrettcolin:
I submitted that using QtCreator's Git plugin I think; I didn't set up an email address, but it didn't stop me from pushing, so I figured maybe it didn't matter. I don't see a blue question mark; the github interface is sort of read only looking for me. It doesn't bother me particularly, though I do apologize for the mess in your nice repository!

dwilliamson:
The one I'm seeing is a gmail.com address. If you add that to your known list of emails it should fix everything from now on https://github.com/settings/emails

PR #39, opened 2015-02-24
Memory override fix on Realloc call

septag:
There is a _realloc_ function call inside the code that breaks the memory overriding feature. And causes apps that override memory to crash.
So I added Realloc callback to settings and replaced the _realloc_ with rmtRealloc to fix the issue.
Sorry for the .gitignore, and Remotery.pro and minor change in remotery.h (_PROFILE_). I didn't know how to remove them in the pull request.

Also when opening the profiler web page, It would cause a 4-byte memory leak on exit. which I haven't figured it out where the origin is. I'll do more test and get back to you later...


Celtoys:
Thanks for this. Can you figure out how to submit just the realloc changes? I could copy/paste your changes myself but I would prefer you got the credit for the changes.

septag:
I have reverted unneeded changes. check it out

Celtoys:
Thanks. Just a heads-up; I'll be deleting the .gitignore file (along with a few others)

PR #48, opened 2015-07-01
add support android ashmem

hjanetzek:
snippet taken from
https://github.com/pelya/android-shmem/blob/master/cutils/ashmem-dev.c

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -293,7 +293,7 @@ static void usTimer_Init(usTimer* timer)
 
         struct timespec tv;
         clock_gettime(CLOCK_REALTIME, &tv);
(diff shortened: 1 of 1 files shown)


dwilliamson:
This is great, thanks. Can you move the ashmem_create_region function into Remotery.c file and remove the need for the header file?

hjanetzek:
Sure, I had it first included but was not sure where to put their copyright notice.

dwilliamson:
Right above the function is fine as that's what I do some some of the other stuff, e.g. https://github.com/Celtoys/Remotery/blob/master/lib/Remotery.c#L934

hjanetzek:
Thanks. Great tool btw! :)

PR #55, opened 2015-07-23
Slightly nicer default thread naming as ThreadN where N is hex thread number. Fixes #40

### LINKED ISSUE #40: Better thread naming
Just a suggestion ...
Instead of hashing _ThreadSampler_ pointers for thread names. It would make more sense to name threads by their IDs. something like :

```
static rmtU32 Thread_GetId()
{
#if defined(RMT_PLATFORM_WINDOWS)

    return GetCurrentThreadId();

#elif defined(RMT_PLATFORM_POSIX)

    return (rmtU32)pthread_self();

#endif
}
```

And inside `ThreadSampler_Constructor` instead of Base64_Encode line :
```
 snprintf(thread_sampler->name, sizeof(thread_sampler->name), "Thread #%d", Thread_GetId());
```

I don't know about windows, but there is also a thread naming feature in posix :
http://man7.org/linux/man-pages/man3/pthread_setname_np.3.html   
You can use that to fetch default thread name (if the user has set it before). If the name is empty, use the Thread ID as the name, if not use the name set by user.

dougbinks:
I've not completely followed the prior string function source code method as this doesn't return an error, however the actual calls don't seem to use that so I think this is sufficient.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -214,7 +214,7 @@ rmtS64 maxS64(rmtS64 a, rmtS64 b)
 // Memory management functions
 static void* rmtMalloc( rmtU32 size )
 {
-	return g_Settings.malloc( g_Settings.mm_context, size );
+    return g_Settings.malloc( g_Settings.mm_context, size );
 }
 
 static void* rmtRealloc( void* ptr, rmtU32 size)
@@ -224,7 +224,7 @@ static void* rmtRealloc( void* ptr, rmtU32 size)
 
 static void rmtFree( void* ptr )
 {
-	g_Settings.free( g_Settings.mm_context, ptr );
+    g_Settings.free( g_Settings.mm_context, ptr );
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dwilliamson:
Thanks. Can you replace tabs with spaces before I pull as I'll end up doing it myself anyway?

dougbinks:
Sure I'll make that change. The only reason I was in tab mode was that the last project I contributed to asked me to change spaces to tabs. The SPACE TAB war continues!

dougbinks:
Done. I also caught some commented out debug code I'd not removed, and a couple of tabs that were already there. For anyone watching this the quick way to do this in VS is select all, Edit->Advanced->Untabify Selected Lines.

dwilliamson:
Yeah, I don't really care about tabs vs. spaces (Remotery is spaces, my game is tabs). I just like to be able to browse the code without spacing problems on Github.

dougbinks:
Thanks. Why on earth github code view isn't designed to work well with mixed tab/spaces I don't know. Would help a lot.

PR #60, opened 2015-08-05
Fix building on Linux with RMT_USE_OPENGL

### LINKED ISSUE #59: Doesn't build on Linux with RTM_USE_OPENGL
It doesn't build out of the box, I had to do a few changes in Remotery.c:
* around Line 5348 it'll explode because `WINGDIAPI` is unknown. so it should really be an `#elif defined(_WIN32)` case + there needs to be a new #else case with `#define GLAPI extern` or something like that.
* `glXGetProcAddressARB` couldn't be found. `#include <gl/glx.h>` create conflicts with your own GLuint typedefs etc, just adding `extern void* glXGetProcAddressARB(const GLubyte*);` made it work.. not pretty, though

And sometimes I get segfaults on startup
```
Program received signal SIGSEGV, Segmentation fault.
Remotery_Destructor (rmt=0xe78750) at ../Libs/Remotery/Remotery.c:4182
4182	        Delete(OpenGL, rmt->opengl);
```
can't reproduce that reliably though and last time I forgot to get a backtrace :-/

Anyway, thanks for this awesome tool, I really enjoy using it and it makes profiling performance problems so much easier :-)

DanielGibson:
Because GL/glx.h isn't #included, it doesn't know about
glXGetProcAddressARB(). Including it causes conflicts with other
definitions like the GLint typedefs, so I just added a declaration.

For the same reason (system GL headers not included). GLAPI isn't known,
so it tried to use WINGDIAPI which is windows-specific, so fall back to
extern on non-_WIN32.

Fix #59

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -5345,8 +5345,10 @@ RMT_API void _rmt_EndD3D11Sample(void)
 #ifndef GLAPI
 #  if defined(__MINGW32__) || defined(__CYGWIN__)
 #    define GLAPI extern
-#  else
+#  elif defined (_WIN32)
 #    define GLAPI WINGDIAPI
+#  else
+#    define GLAPI extern
 #  endif
 #endif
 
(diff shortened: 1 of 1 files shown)


dwilliamson:
thanks

PR #62, opened 2015-11-13
Fixed GCC warnings & errors.

bkaradzic:


lib/Remotery.c
```diff
@@ -508,7 +508,7 @@ static void WriteFence()
 
 // Get a shared value with acquire semantics, ensuring the read is complete
 // before the function returns.
-static void* LoadAcquire(void* volatile const* addr)
+void* LoadAcquire(void* volatile const* addr)
```
dwilliamson: Why remove the static?

dwilliamson:
Thanks for the fixes. Left just one comment.

bkaradzic:
It's removed because function is not referenced anywhere and it complains that it's static and unused (can't be referenced outside of compilation unit).

dwilliamson:
OK, just pulled but getting errors for CRTMalloc/Free/Realloc when compiling in C mode. Will fix.

bkaradzic:
Oh sorry, I was just building with C++.

PR #66, opened 2015-11-15
Support for dynamic sample names.

andr3wmac:
I needed support for dynamic names for profile samples. Didn't seem like there was a way to do this already.


dwilliamson:
Worth bearing in mind that passing NULL as the second parameter is the intended use here but it does mean the name hash will be recalculated on every call.

You may find at a later date that caching your own name hash alongside your dynamic name gives you better performance.

PR #70, opened 2015-11-23
Use prctl with PR_GET_NAME to get the thread name in Linux.

JodiTheTigger:
You can now use Remotery to get and set thread names in linux.

These names can be viewed on the Remotery profiler webpage, as well as Linux system inspection tools (for example, htop).

Due to system implementation, the name is restricted to 15 characters, plus the null terminating character \0.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -4386,8 +4386,8 @@ static void SetDebuggerThreadName(const char* name)
     #ifdef RMT_PLATFORM_POSIX
         // pthread_setname_np is a non-standard GNU extension.
         char name_clamp[16];
-        strncpy(name_clamp, name, 15);
-        name_clamp[15] = 0;
+        name_clamp[0] = 0;
+        strncat_s(name_clamp, sizeof(name_clamp), name, 15);
         prctl(PR_SET_NAME,name_clamp,0,0,0);
     #endif
 }
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -115,6 +115,7 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
         #include <string.h>
         #include <sys/socket.h>
         #include <sys/mman.h>
+        #include <sys/prctl.h>
         #include <netinet/in.h>
         #include <fcntl.h>
         #include <errno.h>
@@ -3650,6 +3651,15 @@ typedef struct ThreadSampler
 
 } ThreadSampler;
 
+static void GetDebuggerThreadName(char* name_max_16_bytes)
+{
+    name_max_16_bytes[0] = 0;
+    #ifdef RMT_PLATFORM_POSIX
+        // pthread_getname_np is a non-standard GNU extension.
+        prctl(PR_GET_NAME,name_max_16_bytes,0,0,0);
+    #endif
+}
+
 static rmtS32 countThreads = 0;
 
 static rmtError ThreadSampler_Constructor(ThreadSampler* thread_sampler)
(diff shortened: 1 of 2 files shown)


dwilliamson:
Thank you, on holiday right now so will review/pull this in a few days.

JodiTheTigger:
Ok cool. Enjoy your holiday!

lib/Remotery.c
```diff
@@ -4369,6 +4382,14 @@ static void SetDebuggerThreadName(const char* name)
     #else
         RMT_UNREFERENCED_PARAMETER(name);
     #endif
+
+    #ifdef RMT_PLATFORM_POSIX
+        // pthread_setname_np is a non-standard GNU extension.
+        char name_clamp[16];
+        strncpy(name_clamp, name, 15);
```
dwilliamson: strncpy is unsafe here. Can you set `name_clamp[0]=0` and call `strncat_s` instead?

lib/Remotery.c
```diff
@@ -3665,9 +3675,12 @@ static rmtError ThreadSampler_Constructor(ThreadSampler* thread_sampler)
     thread_sampler->next = NULL;
 
     // Set the initial name to Thread0 etc.
-    thread_sampler->name[0] = 0;
-    strncat_s(thread_sampler->name, sizeof(thread_sampler->name), "Thread", 6);
-    itoahex_s(thread_sampler->name + 6, sizeof(thread_sampler->name) - 6, AtomicAdd(&countThreads, 1));
+    GetDebuggerThreadName(thread_sampler->name);
```
dwilliamson: What's the reason for doing this? Do the initial `PR_GET_NAME` names returned make more sense than `Thread0` etc. on POSIX?

JodiTheTigger:
Yes. I'm maintaining an app with over 33 threads (ugh). And to mange it well, I have named each thread.

When doing profiling work, it's vital to know what thread is hungry at a glance, since there is no way to easily correlate the name `ThreadX `with the actual thread using it.

lib/Remotery.c
JodiTheTigger: Isn't strncat_s a C11 feature? Could I just put in a null check on `name` instead?

*looks at code*

No wait. I see what you did there :-)

Sure.

lib/Remotery.c
JodiTheTigger: Yes. I'm maintaining an app with over 33 threads (ugh). And to mange it well, I have named each thread. I do this at thread creation so they are set before this code in Remotery gets called.

When doing profiling work, it's vital to know what thread is hungry at a glance, since there is no way to easily correlate the name ThreadXwith the actual thread using it.

JodiTheTigger: Hmm, thinking about this, I've spotted a bug.

By default the thread name inherits the program name, truncated to 15 characters (http://man7.org/linux/man-pages/man3/pthread_setname_np.3.html).

In my code, I set the thread name when creating my threads, but others might not. Therefore all they'll see is many threads with their program name (since thread_sampler->name will never be zero length). In this case the original behaviour would be preferred.

Would it be ok if I made this feature a compile time optional feature, that is off my default?

JodiTheTigger: Ok, I pushed in changes. 
1) Made posix thread naming optional.
2) Merged GetDebuggerThreadName() into ThreadSampler_Constructor().

dwilliamson:
Perfect!

JodiTheTigger:
Bah, I just deleted my branch when I realised I didn't update the readme :-(

PR #77, opened 2016-03-15
Warnings at level 4 on Microsoft Visual Studio 2013 were being created.

paulecoyote:
These are now either being supressed or have been fixed.
This is useful for projects who have had "Warnings are errors" set for level 4.

There's probably better ways of doing this, though this is a good starting point.

Included CMakeLists.txt that generates a Visual Studio project setup in a similar way to the one I was trying to include Remotery in.


lib/Remotery.c
```diff
@@ -2366,12 +2373,19 @@ typedef struct
...
+#endif
     union
     {
         rmtU8 data_mask[4];
         rmtU32 data_mask_u32;
     };
+#ifdef _MSC_VER
+    #pragma warning(pop)
+#endif
 } WebSocket;
```
dwilliamson: Can the union be named instead?

lib/Remotery.c
```diff
@@ -3385,6 +3399,7 @@ static rmtError Sample_Constructor(Sample* sample)
 
 static void Sample_Destructor(Sample* sample)
 {
+    assert(sample == 0 || sample != 0);	// To fool warning C4100 on warning level 4
     RMT_UNREFERENCED_PARAMETER(sample);
 }
```
dwilliamson: This really needs to be solving the fact that `RMT_UNREFERENCED_PARAMETER` isn't working on VS2013

sample/sample.c
```diff
@@ -14,20 +16,30 @@ double delay() {
...
+int sig = 0;
+
+/// Allow to close cleanly with ctrl + c
+void sigintHandler(int sig_num) {
+    sig = sig_num;
+    printf("Interrupted\n");
+}
+
+int main( ) {
+    signal(SIGINT, sigintHandler);
```
dwilliamson: This will fail to build on Windows and other platforms that don't support POSIX signals.

dwilliamson:
Thanks for your PR, Paul. I've left a few comments on what needs to be done. I also won't be able to accept changes with build scripts in them (e.g. CMake).

If you haven't got the time to do this then don't worry, thanks for making me aware of the problems.

sample/sample.c
paulecoyote: This is building on Windows in Visual Studio - I confess I have not tried it on other platforms yet.

lib/Remotery.c
paulecoyote: I could move it in to the RMT_UNREFERENCED_PARAMETER define as "if MIcrosoft then use the assert" ?

lib/Remotery.c
paulecoyote: Not without breaking things.  For example if I name it "u" then you have to reference data_mask via u.

dwilliamson: Yeah, that's exactly what I meant. Name the union `data` and then all variable references change to `data.mask` and `data.mask_u32`. Breaking changes within the .c file are fine.

lib/Remotery.c
dwilliamson: Maybe that would work!

sample/sample.c
dwilliamson: Well, I never! Yup, that works just fine. I can pull this bit.

paulecoyote:
@dwilliamson I'll rebase this before we're done - just want to make sure you have no further feedback.

paulecoyote:
@dwilliamson I can remove the build script (that was really just to demo the problem anyway)
Be nice if we could tuck it away in a script / tools / something folder though - but hey your call - it's trivial to recreate.

dwilliamson:
Definitely can't have any build scripts as it's the first thing that puts me off a project whenever I visit its page. Repo needs to be clean from anything like that.

I am actually thinking of pulling the sample out and making a separate RemoterySamples repo where a few build scripts can be housed.

paulecoyote:
Oh yeah, I can see that RE: Sample.  It's such a small repo already though - think that would only be worth the effort if you had more than one sample.

paulecoyote:
@dwilliamson so that's removed and at the union is named.

dwilliamson:
Perfect, thanks Paul.

I'll pull this later when I can test it.

Now I've seen somebody else disabled warnings in `TCPSocket_PollStatus` - will have to figure out why!

PR #84, opened 2016-05-27
Avoid asserting when the WebSocket server could not be created

### LINKED ISSUE #83: Assertion firing when port already occupied
Whenever binding the TCP port for the Websocket server fails, the `Remotery_Constructor` returns w/ an error code before having set the `g_Remotery`/`g_RemoteryCreated` globals.

As the `EndNew` macro calls `Remotery_Destructor` automatically on error, it will fire the assertions that begin that function:

```
assert(g_RemoteryCreated == RMT_TRUE)
```

My proposed fix would be to move the assertions into `rmt_DestroyGlobalInstance` and in the destructor erase g_Remotery if and only if g_RemoteryCreated is true

nil-ableton:
This fixes #83

If for instance the port is already occupied, `Remotery_Constructor`
will return with an error before g_Remotery and g_RemoteryCreated could
be set.

This situation triggers an assertion immediately in `Remotery_Destructor`
since it expects these globals to have been set.

This fixes it by asserting only when coming from the intended API entry
point, `rmt_DestroyGlobalInstance` which is where the assertion would be
most useful.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -4233,11 +4233,11 @@ static void Remotery_Destructor(Remotery* rmt)
     // Join the remotery thread before clearing the global object as the thread is profiling itself
     Delete(Thread, rmt->thread);
 
-    // Ensure this is the module that created it
-    assert(g_RemoteryCreated == RMT_TRUE);
-    assert(g_Remotery == rmt);
-    g_Remotery = NULL;
-    g_RemoteryCreated = RMT_FALSE;
+    if (g_RemoteryCreated)
+    {
+      g_Remotery = NULL;
+      g_RemoteryCreated = RMT_FALSE;
+    }
 
     #if RMT_USE_D3D11
         Delete(D3D11, rmt->d3d11);
@@ -4393,6 +4393,9 @@ RMT_API rmtError _rmt_CreateGlobalInstance(Remotery** remotery)
 
 RMT_API void _rmt_DestroyGlobalInstance(Remotery* remotery)
 {
+    // Ensure this is the module that created it
+    assert(g_RemoteryCreated == RMT_TRUE);
+    assert(g_Remotery == remotery);
     Delete(Remotery, remotery);
 }
 


lib/Remotery.c
```diff
@@ -4233,11 +4233,11 @@ static void Remotery_Destructor(Remotery* rmt)
...
-    // Ensure this is the module that created it
-    assert(g_RemoteryCreated == RMT_TRUE);
-    assert(g_Remotery == rmt);
-    g_Remotery = NULL;
-    g_RemoteryCreated = RMT_FALSE;
+    if (g_RemoteryCreated)
+    {
+      g_Remotery = NULL;
+      g_RemoteryCreated = RMT_FALSE;
+    }
```
dwilliamson: I'm looking at the code and it doesn't look like the `if (g_RemoteryCreated)` check is needed:

- If `g_RemoteryCreated == RMT_FALSE` then there is no harm in setting it to `RMT_FALSE` anyway.
- When `g_RemoteryCreated == RM_FALSE` and `g_Remotery` is a valid pointer (from `rmt_SetGlobalInstance`), the destructor will never be called as the assert below would trap it before then.

Does that make sense?

nil-ableton: Yes that makes sense.

I wrote it in this way to mimick the New() .. Delete(...) pattern which is used for other aspects of the remotery instance to deal with conditional destruction of a constructed part.

I.e. the pattern that Delete(...) follows of destructing iff the object is not null

dwilliamson: Right. The "assign to global pointer" pattern is broken in there anyway and doesn't really adhere to the New/Delete lifetime pattern (`Remotery_ThreadMain` needs to be created outside the Remotery constructor). But there is value in bringing it closer.

dwilliamson:
Thanks for fixing this.

nil-ableton:
Thanks your quick feedback!

PR #86, opened 2016-05-31
Disable SIGPIPE signals when the client disconnects

nil-ableton:
When running Remotery on MacOS I see SIGPIPE signals
emitted when playing around with the browser based client
(refreshing it for instance)

When this signal is not handled by an application this leads
to a crash.

Solution:

- BSD, we set the `SO_NOSIGPIPE` flag using `setsockopt`
- Linux, we pass the `MSG_NOSIGNAL` flag to `send`

Tested on MacOSX 10.11

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -1830,6 +1830,18 @@ static rmtError TCPSocket_AcceptConnection(TCPSocket* tcp_socket, TCPSocket** cl
     if (s == SOCKET_ERROR)
         return RMT_ERROR_SOCKET_ACCEPT_FAIL;
 
+#ifdef SO_NOSIGPIPE
+    // On POSIX systems, send() may send a SIGPIPE signal when writing to an
+    // already closed connection. By setting this option, we prevent the
+    // signal from being emitted and send will instead return an error and set
+    // errno to EPIPE.
+    //
+    // This is supported on BSD platforms and not on Linux.
+    {
+        int flag = 1;
+        setsockopt(s, SOL_SOCKET, SO_NOSIGPIPE, &flag, sizeof(flag));
+    }
+#endif
     // Create a client socket for the new connection
     assert(client_socket != NULL);
     New_0(TCPSocket, *client_socket);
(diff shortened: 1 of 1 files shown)


nil-ableton:
References:
For linux, http://linux.die.net/man/2/send
For BSD, https://www.freebsd.org/cgi/man.cgi?query=setsockopt&sektion=2

lib/Remotery.c
```diff
@@ -1885,7 +1897,13 @@ static rmtError TCPSocket_Send(TCPSocket* tcp_socket, const void* data, rmtU32 l
...
     {
         // Attempt to send the remaining chunk of data
-        int bytes_sent = (int)send(tcp_socket->socket, cur_data, (int)(end_data - cur_data), 0);
+        int send_flags = 0;
+#ifdef MSG_NOSIGNAL
+        // On Linux this prevents send from emitting a SIGPIPE signal
+        // Equivalent on BSD to the SO_NOSIGPIPE option.
+        send_flags = MSG_NOSIGNAL;
+#endif
+        int bytes_sent = (int)send(tcp_socket->socket, cur_data, (int)(end_data - cur_data), send_flags);
```
dwilliamson: This won't compile in C mode.

lib/Remotery.c
```diff
@@ -1885,7 +1897,13 @@ static rmtError TCPSocket_Send(TCPSocket* tcp_socket, const void* data, rmtU32 l
     while (cur_data < end_data)
     {
         // Attempt to send the remaining chunk of data
-        int bytes_sent = (int)send(tcp_socket->socket, cur_data, (int)(end_data - cur_data), 0);
+        int send_flags = 0;
+#ifdef MSG_NOSIGNAL
+        // On Linux this prevents send from emitting a SIGPIPE signal
+        // Equivalent on BSD to the SO_NOSIGPIPE option.
+        send_flags = MSG_NOSIGNAL;
```
dwilliamson: This won't compile in C mode.

dwilliamson:
Thanks, again. Can you adjust your tabs to 4xSpaces?

nil-ableton:
I've corrected the spacing and moved the variable declarations at the top of the scope. Good catch :)

dwilliamson:
Good job. Thanks for the thorough explanation.

PR #87, opened 2016-06-07
Some improvements/bugfixes + rmt_GetErrorString()

DanielGibson:
* Set SO_REUSEADDR for server socket, so binding doesn't fail right after restarting your application
* In Remotery_Constructor() make sure all pointers are initialized to NULL
  * it could happen that `rmt->opengl` was set to garbage if Remotery_Constructor() returned an error early
* Implemented `rmt_GetErrorString(rmtError errcode)`

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -4198,6 +4198,23 @@ static rmtError Remotery_Constructor(Remotery* rmt)
     rmt->json_buf = NULL;
     rmt->thread = NULL;
 
+    #if RMT_USE_CUDA
+        rmt->cuda.CtxSetCurrent = NULL;
+        rmt->cuda.EventCreate = NULL;
+        rmt->cuda.EventDestroy = NULL;
+        rmt->cuda.EventElapsedTime = NULL;
+        rmt->cuda.EventQuery = NULL;
+        rmt->cuda.EventRecord = NULL;
+    #endif
+
+    #if RMT_USE_D3D11
+        rmt->d3d11 = NULL;
+    #endif
+
+    #if RMT_USE_OPENGL
+        rmt->opengl = NULL;
+    #endif
+
     // Kick-off the timer
     usTimer_Init(&rmt->timer);
 
@@ -4221,26 +4238,13 @@ static rmtError Remotery_Constructor(Remotery* rmt)
     if (error != RMT_ERROR_NONE)
         return error;
 
-    #if RMT_USE_CUDA
-
-        rmt->cuda.CtxSetCurrent = NULL;
-        rmt->cuda.EventCreate = NULL;
-        rmt->cuda.EventDestroy = NULL;
-        rmt->cuda.EventElapsedTime = NULL;
-        rmt->cuda.EventQuery = NULL;
-        rmt->cuda.EventRecord = NULL;
-
-    #endif
-
     #if RMT_USE_D3D11
-        rmt->d3d11 = NULL;
         error = D3D11_Create(&rmt->d3d11);
         if (error != RMT_ERROR_NONE)
             return error;
     #endif
 
     #if RMT_USE_OPENGL
-        rmt->opengl = NULL;
         error = OpenGL_Create(&rmt->opengl);
         if (error != RMT_ERROR_NONE)
             return error;
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -1699,6 +1699,7 @@ static rmtError TCPSocket_RunServer(TCPSocket* tcp_socket, rmtU16 port, rmtBool 
 {
     SOCKET s = INVALID_SOCKET;
     struct sockaddr_in sin;
+    int enable = 1;
     #ifdef RMT_PLATFORM_WINDOWS
         u_long nonblock = 1;
     #endif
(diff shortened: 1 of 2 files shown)


dwilliamson:
Thanks for opening this.

I'm on contract right now so I'll be a little slow getting through it.

I'll add some comments to the PR.

lib/Remotery.c
```diff
@@ -1711,6 +1711,22 @@ static rmtError TCPSocket_RunServer(TCPSocket* tcp_socket, rmtU16 port, rmtBool
...
+        setsockopt(s, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(int));
+    #elif defined(RMT_PLATFORM_WINDOWS)
+        // windows also needs SO_EXCLUSEIVEADDRUSE,
+        // see http://www.andy-pearce.com/blog/posts/2013/Feb/so_reuseaddr-on-windows/
+        int enable = 1;
+        setsockopt(sd, SOL_SOCKET, SO_REUSEADDR, (char *)&enable, sizeof(enable));
+        enable = 1;
+        setsockopt(sd, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char *)&enable, sizeof(enable));
+    #endif
+
```
dwilliamson: Good catch. This won't compile in C mode though.

lib/Remotery.c
```diff
@@ -4674,6 +4690,84 @@ RMT_API void _rmt_EndCPUSample(void)
...
+{
+    int i;
+    for(i=0; i<sizeof(rmtErrorMappings)/sizeof(rmtErrorMappings[0]); ++i)
+    {
+        if(rmtErrorMappings[i].code == errcode)
+            return rmtErrorMappings[i].str;
+    }
+
+    return "Unknown Error/Invalid Error Code";
+}
```
dwilliamson: Right now I'm not inclined to take any error string modifications like this as I think the whole API is broken.

I want to remove enums from the public API and have all functions return an opaque integer error (0=success). This could then be passed to rmt_GetErrorString.

However if you fancy a crack at this... :)

lib/Remotery.c
```diff
@@ -4176,12 +4192,12 @@ static rmtError Remotery_Constructor(Remotery* rmt)
     assert(rmt != NULL);
 
     // Set default state
-    rmt->server = NULL;
+
+    // set all pointers to NULL, incl. rmt->opengl etc,
+    // so they're initialized even if we error out early
+    memset(rmt, 0, sizeof(*rmt));
+
```
dwilliamson: I know it's a pain but could you just copy the null pointers set below that up to here and rewarp them with their defines?

I don't want to set a precedent for using memset to set default state on an object in its constructor.

lib/Remotery.c
DanielGibson: not C89 at least.. I'll just move  the `int enable` to the top of the function

lib/Remotery.c
DanielGibson: you you basically want to replace rmtError with int in the public API and move the enum to Remotery.c?

lib/Remotery.c
DanielGibson: ok

lib/Remotery.h
```diff
@@ -86,10 +86,10 @@ documented just below this comment.
 // Compiler identification
 #if defined(_MSC_VER)
     #define RMT_COMPILER_MSVC
-#elif defined(__GNUC__)
-    #define RMT_COMPILER_GNUC
 #elif defined(__clang__)
     #define RMT_COMPILER_CLANG
+#elif defined(__GNUC__)
```
DanielGibson: Depending on the usecase of RMT_COMPILER_* (none so far...), maybe both RMT_COMPILER_GNUC and RMT_COMPILER_CLANG should be defined for clang, as clang supports (almost?) all GCC features?

DanielGibson:
Thanks for the feedback!
I did the changes according to your comments, except for rmt_GetErrorString() as that seems to be at least a bit more work - should I put that in a separate pull request?

lib/Remotery.c
dwilliamson: Yes. Or something like that. I've only thought about the public API and not how it's implemented. Open to options.

lib/Remotery.h
dwilliamson: I didn't know that was there! I'd just remove them all to be fair. They're not used and just cluttering up the API.

dwilliamson:
If you separate out the error string stuff then I'll just go ahead and accept this. The error codes can be handled later.

DanielGibson:
ok, squashed the commits and removed the rmt_GetErrorString() commit, should be ok now

dwilliamson:
Thanks!

lib/Remotery.c
uucidl: In a multiprocess environment, I've been using the remotery error code `RMT_ERROR_SOCKET_BIND_FAIL`

When creating my Remotery instances, I check whether this occurs and if it does not I bump the port number by one and try again.

This let me monitor the two processes that are spawned by the program.

.. Checking against an error string would be doable but cumbersome as I wouldn't be sure anymore what kind of error occurred.

dwilliamson: Thanks for sharing your use-case. It makes sense but it's really annoying!
Maybe what needs to happen is that the error codes returned to the caller need to be pruned.

DanielGibson: yep, that's a usecase I also thought about when looking at the error code and wondering which of them one would actually react to

uucidl: Pruning the error codes to a few higher level actionable ones would certainly work

PR #92, opened 2016-06-22
Removed 'defined but not used' warnings 

JodiTheTigger:
Removed 'defined but not used' warnings when building with RMT_USE_OPENGL set to 0.


dwilliamson:
Apologies, I'm behind on Remotery maintenance right now but will review these within a few days.

PR #96, opened 2016-08-13
Fixed some compile errors and warnings given by MSVC 2015

Skylark13:
- missing stdlib.h for strtoul
- removed unused variable
- comparison of rmtU32 to NULL --> 0
- ::bind gave an error on the first ':', not sure what's going on there or if this will break other platforms/compilers
- needed to disable another warning for the getProcAddress stuff

Note: suggestions welcome for the ::bind() one...

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -101,6 +101,7 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
         #ifndef __MINGW32__
             #include <intrin.h>
         #endif
+        #include <stdlib.h>
         #undef min
         #undef max
     #endif
@@ -1892,7 +1893,6 @@ static rmtError HashTable_Insert(HashTable* table, rmtU32 key, rmtU32 value)
 static rmtError HashTable_Resize(HashTable* table)
 {
     rmtU32 old_max_nb_slots = table->max_nb_slots;
-    rmtU32 new_nb_occupied_slots = 0;
     HashSlot* new_slots = NULL;
     HashSlot* old_slots = table->slots;
 
@@ -1918,7 +1918,7 @@ static rmtError HashTable_Resize(HashTable* table)
     for (rmtU32 i = 0; i < old_max_nb_slots; i++)
     {
         HashSlot* slot = old_slots + i;
-        if (slot->key != NULL)
+        if (slot->key != 0)
             HashTable_Insert(table, slot->key, slot->key);
     }
 
@@ -2143,7 +2143,7 @@ static rmtError TCPSocket_RunServer(TCPSocket* tcp_socket, rmtU16 port, rmtBool 
     sin.sin_family = AF_INET;
     sin.sin_addr.s_addr = htonl(limit_connections_to_localhost ? INADDR_LOOPBACK : INADDR_ANY);
     sin.sin_port = htons(port);
-    if (::bind(s, (struct sockaddr*)&sin, sizeof(sin)) == SOCKET_ERROR)
+    if (bind(s, (struct sockaddr*)&sin, sizeof(sin)) == SOCKET_ERROR)
         return RMT_ERROR_SOCKET_BIND_FAIL;
 
     // Connection is valid, remaining code is socket state modification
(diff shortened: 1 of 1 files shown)


zao:
With the file type being C, isn't `::` for global namespace a C++ism completely out of place?

Celtoys:
Yes, :: should not be there and is a side-effect left over from a merge elsewhere.

Thanks for the PR.

Celtoys:
Is this compiled as C++? Is there no other way around it other than disabling warnings?

Skylark13:
You're right, I compile my whole project as C++, and Remotery was integrated by just putting the header and C file into it. I guess you mean for the "#pragma warning(disable:4055)" -- I'm not sure there's another way to fix that warning that would be cross-platform.

The thing is, on Windows GetProcAddress returns a function pointer, and VS warns when casting void* (a data pointer) to a function pointer. So for projects that compile with warnings-as-errors, that's a problem.

I just tried this, it also removes the warnings without needing to disable them, but it has the downside of changing the return type of rmtGetProcAddress:

```C
typedef INT_PTR(WINAPI *PROC)();

static PROC rmtGetProcAddress(void* handle, const char* symbol)
{
    #if defined(RMT_PLATFORM_WINDOWS)
        return GetProcAddress((HMODULE)handle, (LPCSTR)symbol);
    #elif defined(RMT_PLATFORM_POSIX)
        return dlsym(handle, symbol);
    #else
        return NULL;
    #endif
}

static PROC rmtglGetProcAddress(OpenGL* opengl, const char* symbol)
{
    assert(opengl != NULL);

    #if defined(RMT_PLATFORM_WINDOWS)
    {
        // Get OpenGL extension-loading function for each call
        typedef PROC (*wglGetProcAddressFn)(LPCSTR);
        {
            wglGetProcAddressFn wglGetProcAddress = (wglGetProcAddressFn)rmtGetProcAddress(opengl->dll_handle, "wglGetProcAddress");
            if (wglGetProcAddress != NULL)
                return wglGetProcAddress(symbol);
        }
    }
    #elif defined(__APPLE__) && !defined(GLEW_APPLE_GLX)
        return NSGLGetProcAddress((const GLubyte*)symbol);
    #elif defined(RMT_PLATFORM_LINUX)
        return glXGetProcAddressARB((const GLubyte*)symbol);
    #endif

    return NULL;
}
```

I guess if for other platforms you typedef void* PROC; that might work and behave the same as the old code. I can try to amend the PR if you want.

dwilliamson:
Instead of introducing your own type, does using the return type `FARPROC` work? `GetProcAddress` and `dlsym` have different return types so it makes sense that `rmtGetProcAddress` vary its return type based on platform, too.

Skylark13:
Sorry for taking so long to reply.

Yes, looks like using FARPROC works, so you can remove the typedef and replace PROC with FARPROC.

~~~C++
static FARPROC rmtGetProcAddress(void* handle, const char* symbol)
{
    #if defined(RMT_PLATFORM_WINDOWS)
        return GetProcAddress((HMODULE)handle, (LPCSTR)symbol);
    #elif defined(RMT_PLATFORM_POSIX)
        return dlsym(handle, symbol);
    #else
        return NULL;
    #endif
}

static FARPROC rmtglGetProcAddress(OpenGL* opengl, const char* symbol)
{
    assert(opengl != NULL);

    #if defined(RMT_PLATFORM_WINDOWS)
    {
        // Get OpenGL extension-loading function for each call
        typedef FARPROC (*wglGetProcAddressFn)(LPCSTR);
        {
            wglGetProcAddressFn wglGetProcAddress = (wglGetProcAddressFn)rmtGetProcAddress(opengl->dll_handle, "wglGetProcAddress");
            if (wglGetProcAddress != NULL)
                return wglGetProcAddress(symbol);
        }
    }
    #elif defined(__APPLE__) && !defined(GLEW_APPLE_GLX)
        return NSGLGetProcAddress((const GLubyte*)symbol);
    #elif defined(RMT_PLATFORM_LINUX)
        return glXGetProcAddressARB((const GLubyte*)symbol);
    #endif

    return NULL;
}
~~~

This compiles without warnings for me.

PR #98, opened 2016-09-05
Removed Linux unused variables/functions warnings.

JodiTheTigger:
Fixed build warnings for Linux, which was a combination of
ununsed variables, set but unused variables, unused functions,
illegal C++ :: specifier, C89 illegal syntax and casting warnings.

PR #104, opened 2016-10-26
Fixed a small error that should not appear in C code.

Deathsbreed:
I know this has been solved in other pull requests (namely #98 and #96), however those are grouped together with other fixes. This fixes the `::bind` error that we've all been getting for once and it'll merge just fine. Also, even if you mark this as a duplicate, at least to the favor of merging one of the other pull requests.


Celtoys:
Thanks

PR #105, opened 2016-10-29
Fixed the comparison warning.

Deathsbreed:
Here's the second fix. There's this annoying warning that turns into an error when you have a strict compiler. Someone else tried fixing it, but I don't believe they did it correctly. We're not necessarily trying to see if `slot->key` is `0`, but rather if the pointer (`slot`) exists to begin with, 'cause since `slot->key` is stack allocated, if `slot` exists then so will `key`. So, we check if `slot` exists.


Celtoys:
Thanks

PR #115, opened 2017-03-06
add support for FreeBSD

zholos:
Here are some tweaks to make Remotery compile and work on FreeBSD.

FreeBSD's APIs are pretty similar to Linux, so I just based it on RMT_PLATFORM_LINUX.

The differences are:

* No malloc.h

* CLOCKS_PER_SEC is only 128. The comment next to msTimer_Get() seems to imply this resolution is sufficient, so I just avoided the division by zero in `time / (CLOCKS_PER_SEC / 1000)`. Another option is getrusage().

* No /dev/shm

* No prctl(). Thread name can be set with pthread_set_name_np() (and gdb sees it) but there's no function to get the name.

I tested profiling CPU and OpenGL with multiple threads. Seems to work fine.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -92,7 +92,9 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
         #include <mach/mach.h>
         #include <sys/time.h>
     #else
-        #include <malloc.h>
+        #ifndef __FreeBSD__
+            #include <malloc.h>
+        #endif
     #endif
 
     #include <assert.h>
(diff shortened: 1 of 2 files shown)


dwilliamson:
Great work, thanks for the PR.

barrettcolin:
Unfortunately "#if CLOCKS_PER_SEC < 1000 // 128 on FreeBSD" doesn't compile on Linux (Fedora 23, gcc 5.3.1), on account of "#define CLOCKS_PER_SEC  ((clock_t) 1000000)"

dwilliamson:
Ouch. How does the divide work in that case? Isn't `clock_t` a `long int` in glibc?

barrettcolin:
The preprocessor doesn't know what to do with (clock_t) in the comparison, I expect. If I look on Raspbian, there it's "#define CLOCKS_PER_SEC 1000000l". Don't know why it is the way it is on Fedora

dwilliamson:
oh doh of corse

dwilliamson:
Right, might as well just be explicit about the platform https://github.com/Celtoys/Remotery/commit/6a1a5f80ef7d70509d965d16945cf719d04d152c

zholos:
Oops. That way works too. Thanks!

PR #120, opened 2017-09-29
Bugfix: Layout issues in CSS that caused two bugs in the sample windows:

darksylinc:
* The height of the sample windows would not follow the browser zoom
* Scrolling was broken, making navigation through lots of samples very
difficult.
This fixes layout issues with latest version of Firefox, as well as
Chrome.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -1937,7 +1937,7 @@ static rmtError rmtHashTable_Resize(rmtHashTable* table)
     {
         HashSlot* slot = old_slots + i;
         if (slot->key != 0)
-            rmtHashTable_Insert(table, slot->key, slot->key);
+            rmtHashTable_Insert(table, slot->key, slot->value);
     }
 
     free(old_slots);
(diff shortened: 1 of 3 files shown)


darksylinc:
A tour about the changes:

Without "height: inherit", browser would not calculate height correctly and the scroll guide at the right would just not work. It would be there, but you couldn't interact with it. Only way to scroll up or down was via mouse wheel. This was annoying for large number of samples.

The 2nd change, the height was hardcoded to 400px. This worked more or less well for zoom in (it had a few issues, mostly objects at the bottom being hidden because the browser wouldn't let you scroll enough), but it worked like crap when zooming out.
Setting it to "100%" or "inherit" wouldn't work as intended either because there is an element above that was exactly 17px (I think it's the window name, but I'm not sure). Solution was to set the height to "calc( 100% - 17px )" which makes it fit perfectly to the container's height. Now it adapts perfectly to zoom in and out without hiding any element.

darksylinc:
Mmm... It seems GitHub automatically updated this PR.
I wanted to do it in separate PRs.

I added a fix to Remotery.c; as resizing the HashTable would result in crashes. Once you see the change (one line) it becomes pretty obvious:
`rmtHashTable_Insert(table, slot->key, slot->key);`
vs
`rmtHashTable_Insert(table, slot->key, slot->value);`

dwilliamson:
That hash table bug is crazy and would explain some odd name behaviour now and then. Good find.

PR #121, opened 2017-10-01
Added synchronization for GPU timeline relative to CPU's timeline

darksylinc:
Both OpenGL & D3D11 returned timestamps that were not relative to the
same time space the CPU samples were. This made comparison between GPU
and CPU timelines impossible; as each timeline was independent.

This code adds SyncD3D11CpuGpuTimes & SyncOpenGLCpuGpuTimes which stall
rendering, then proceeds to measure the average RTT (Roundtrip Time)
CPU->GPU->CPU and finally calculates the offset in which GPU timestamps
need to be applied to become relative to CPU.

By default Remotery will resync if:
* It detects an invalid sample (GPU sample's timestamp says it happened
before the CPU command was issued)
* D3D11 reported a disjoint
* 30 seconds have elapsed since the last sync

Because resyncs can slowdown application performance or cause stutters,
this behavior can be managed or even completely disabled via the new
macros:
RMT_GPU_CPU_SYNC_NUM_ITERATIONS
RMT_GPU_CPU_SYNC_SECONDS
RMT_D3D11_RESYNC_ON_DISJOINT

Remotery now requires glFinish & glGetInteger64v

These code changes have been tested in both Linux & Windows.

According to my measures, the timelines appear to be syncing very
accurately, as timelines look as expected when enabling VSync, disabling
VSync and keeping a max 3 frames window, and when forcibly causing
stalls. However I cannot say if the timelines are accurate to the
microsecond, but in theory they should be.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -6264,7 +6264,7 @@ static void SyncOpenGLCpuGpuTimes(rmtU64* out_first_timestamp, rmtU64* out_last_
     rmtU64 cpu_time_start = 0;
     rmtU64 cpu_time_stop = 0;
     rmtU64 average_half_RTT = 0; //RTT = Rountrip Time.
-    rmtU64 gpu_base = 0;
+    GLint64 gpu_base = 0;
     int i;
 
     rmtglFinish();
@@ -6289,7 +6289,7 @@ static void SyncOpenGLCpuGpuTimes(rmtU64* out_first_timestamp, rmtU64* out_last_
     // All GPU times are offset from gpu_base, and then taken to
     // the same relative origin CPU timestamps are based on.
     // CPU is in us, we must translate it to ns.
-    *out_first_timestamp = gpu_base - (cpu_time_start + average_half_RTT) * 1000ULL;
+    *out_first_timestamp = (rmtU64)(gpu_base) - (cpu_time_start + average_half_RTT) * 1000ULL;
     *out_last_resync = cpu_time_stop;
 }
 
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -5500,6 +5500,8 @@ typedef struct D3D11
 
     // Mark the first time so that remaining timestamps are offset from this
     rmtU64 first_timestamp;
+    // Last time in us (CPU time, via usTimer_Get) since we last resync'ed CPU & GPU
+    rmtU64 last_resync;
 } D3D11;
 
 
@@ -5520,6 +5522,7 @@ static rmtError D3D11_Create(D3D11** d3d11)
     (*d3d11)->last_error = S_OK;
     (*d3d11)->mq_to_d3d11_main = NULL;
     (*d3d11)->first_timestamp = 0;
+    (*d3d11)->last_resync = 0;
 
     New_1(rmtMessageQueue, (*d3d11)->mq_to_d3d11_main, g_Settings.messageQueueSizeInBytes);
     if (error != RMT_ERROR_NONE)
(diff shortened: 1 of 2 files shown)


darksylinc:
Before this enhancement, GPU and CPU timelines were completely independent. It only worked "more or less" well if you happen to start the API samples and the CPU samples at the exact same time, which is rarely the case because it's common to initialize remotery at startup (thus leaving a huge gap between Remotery's CPU start and the API initialization); or if you were lucky enough the GPU returned timestamps in the same space CPU is in (rarely the case, specially for discrete GPUs)

A timeline like the following was very common:
remotery_old
Notice the GPU is many frames behind the CPU, which does not correspond to what's happening.

After these code enhancements, timelines are sync and now look like this when they're with VSync:
remotery_new_withvsync

And this is how it looks when VSync is off:
remotery_new_novsync

Having the timelines in sync is invaluable for spotting pipeline bubbles, diagnosing frame latency, stalls, pipeline scheduling issues, etc.

darksylinc:
Note: The change in colours (from blue in the old one, to rainbow in the new one) is not part of these code enhancements.
That was just some trick of passing frame number as the name hash, instead of passing a name and relying on remotery to hash it. The "old" code didn't have it because the screenshot was taken with an earlier build that didn't have this "trick".

dwilliamson:
This is amazing, will be reviewing it tonight. In your experience, how much does the round-trip depend on the current GPU load?

darksylinc:
I didn't see the question until now!

I'm not sure I understand the question.

The problem when syncing GPU and CPU is that we don't know when the GPU actually measured time with its own timer.

For example, let's divide time in discrete steps for making it easier:
1. Halt GPU. (i.e. glFinish)
2. CPU measures time. CPU time now = 15 (doesn't matter if it's ns, us, seconds)
3. Issue a GPU query timestamp
4. Wait for the result
5. CPU measures time. CPU time now = 101. GPU time reported = 3546

We know that the GPU measured time somewhere between 15 and 101. If the CPU and GPU are doing nothing else other than taking time, then it is a good a approximation that the GPU took the time around the middle = (15 + 101) / 2 = 58.

Maybe it's a little more biased towards the end, maybe a towards the start. During steps 2 and 5; the CPU sent a command to the driver, the driver set a command to the GPU, the GPU measured time, reported back the results to the driver, and the driver reports back the result to us. All in all, it's a good approximation to think it happened exactly in the middle.

But this approximation only works if we stop both CPU & GPU from doing everything they were doing (hence the calls to glFinish and rmtD3D11Finish; and performing a live spinlock waiting for the results). If the GPU and/or CPU are doing other stuff (like rendering) the bias would be gigantic.
Initially I was doing;
1. CPU measures time. CPU time now = 15 (doesn't matter if it's ns, us, seconds)
2. Issue a GPU query timestamp for regular GPU samples (timestamp begin)
3. Keep rendering
4. Issue a GPU query timestamp for regular GPU samples (timestamp end)
5. Eventually the two GPU queries becomes available
6. CPU measures time. CPU time now = 1001. GPU time reported begin = 3546 end = 4987

Now... when did the GPU measure time? Did it measure 3546 close to 15? did it measure 4987 close to 1001? And how do I know how much time happened between the GPU measured "4987" and we measured "1001" in the CPU?
There's a lot of noise in there. Whatever happens in "Keep rendering" could be something really expensive, or something really cheap.
I tried this approach and the overal CPU/GPU sync worked however I had errors of up to an entire frame, which is a lot.

With this approach the timelines were really well sync'ed as I removed all the noise.

The only drawback of this technique is that the measuring time could change the power profile (because measuring time for syncing is cheaper/more expensive than what the user was rendering so the driver changes the GPU clocks) but it didn't seem to affect my readings. My guess is that:
1. Power profiles didn't actually change
2. If driver is indeed changing power profiles, the timers aren't affected or the change is proportional, or the error is not big enough to notice.
3. If you're really worried about this, it's common sense to use something like D3D12's SetStablePowerState.exe or other driver tweaks to get stable readings. Syncing is not the only readings that are going to be affected by power saving modes.

So that's why I don't understand the question "round-trip depend on the current GPU load?". Since we basically get rid of all previous GPU load in order to measure the rtt.
Based on the debugger values, the RTTs I get in my machine is around 240us for D3D11, but this is in Debug, using the Debug layer. OpenGL in Debug gives me 18us.

PR #124, opened 2017-10-13
! added save / load

mihaisebea:
Hi 
I added a way to serialize and deserialize data.
Since this is my first time writing javascript please have mercy on my poor soul :)

Some things that I'm not sure about 
1. I added the datastream library which seems to have the same Apache license so i guess there are no issues with it !
2. When loading i clear everything and stop the auto connecting and set the status to pause.
- i suppose I should also notify the button to set it's state to checked or pressed?
3. sometimes you load a file and nothing happens ... i guess i'm doing something wrong with the event handler for the input file ... couldn't figure out what's going on ... the debugger doesn't throw any errors :(

I might be doing other stuff wrong ... like not preserving the project's tab / spaces polices or general coding conventions ... and for that I'm sorry :|

Feedback is welcome :)


dwilliamson:
Thanks so much for submitting this!

As it stands it's not something I'd pull over because I want to solve the problem in a more complete way. The problems are:

* The viewer is not the complete story: if the ring buffer on the C side overflows through too much data being sent, samples are dropped.
* Also, the viewer can't hold all samples in memory for big programs so it continually discards older samples.

There is no real way to solve this from the client so it must be done from C. The plan is:

* The C side emits log files in a way similar to Chrome Traces. That is; non-hierarchical so you can incrementally write all data before a crash. Thinking about it, support for Chrome Trace JSON files would be awesome.
* The web viewer can load and provide a small window into those log files, irrespective of how large they are.

I can totally see what you've done being useful on smaller projects but I use Remotery for my clients, that usually have 10s of threads emitting 1000s of samples a frame. I need to preserve all behaviour for further playback (hence why right now I'm actually using Chrome).

I'll keep this open if you'd like to adapt what you've written. Let me know.

neo2buha:
Hi 
i'm not sure i understand the problems ...
- if the ring buffer can overflow can we make it bigger ?  ...  minimize the amount of data we send per sample , compress it (LZ4 ?) so we don't overload the network ( if that is a problem ?)
- for the number of samples I've seen a limit of 10k samples hardcoded somewhere !  
I was actually thinking we could have an option to save them automatically to disk before dropping them ! 

In any case ... from the "Server" side 
- not sure I understand what you mean by emitting ? write them to disk ? send them over the network ?
To me save them locally would be a problem since I usually profile on phones I hate having to connect and extract files ! I don't really see the point in sending them over the network ...we are already sending the samples right ?
- In case ...i think this little library could be used to write chrome trace json
https://github.com/hrydgard/minitrace    
- also I'm not sure what you mean by non hierarchical ... I assume that if a parent scope did not finish we should still send the child samples ? I think I have seen other discussion about this in some other pull requests ...  

- Regarding the webviewer it seems it can load 60Mb files without a breeze ! 
The only issues I'm having with huge number of samples is 
a) the fact that samples window refresh on every sample ... 
- while this is nice I have no practically utility for it ... samples would arrive so fast that it's impossible to view them 
- also for big enough samples this completely hogs the   CPU 
- and for this I actually added an autorefresh option in this very pull request that stops the sample window refreshing on each sample ! 
b) the second problem which for me is huge is the lack of scroll bar or the lack of collapsing blocks 
- but again I'm too noob in javascript to know how to modify this 
- I'm actually planning to do a Client in QT with C++ where i could do this much more easly 
- don't hold your breath for this pull request though ... I'm not sure when I'll have time to implement it and publish it ! 
 Out of curiosity how do you clients work with these limitations ?

In Conclusion ... I have to say that I'm a bit disappointed that my PR didn't get pulled but I get it ! 
I do appreciate you leaving it open maybe somebody will find it useful !
If you can further detail your vision  on how the system should  work on both server and client I will try to find the time to implement them !

dwilliamson:
This was auto-closed when I changed the default branch from `master` to `main`.

However, latest version of Remotery as of today supports emitting the complete network stream of messages to file. Nothing is missed.

These log files can be loaded by the viewer by simply dragging and dropping them into the browser window.

PR #140, opened 2018-08-12
Console fixes

tillrathmann:
I integrated Remotery into my project and made use of the console feature.
* Then I realized that somehow after a while there are no new messages displayed in the Remotery console window. I saw that after 10*1024 characters the viewer crops the beginning of the string. However, this means the size of the string remains the same but the size was the criterion to determine if a html update is needed. So I changed the update trigger.
* During debugging I also noticed that there is a buffer overrun that causes a stack corruption when calling `rmt_LogText()` with \ or " at the "right" location near line_buffer end. Calling `rmt_LogText("\\\\\\.......<many \\'s here>....\\\\")` will trigger this crash for example. In line 5214 `prev_offset` will be `sizeof(line_buffer) - 4` at some point. Then is will be increased by two (because of the \ --> \\\\ conversion) and in the next iteration `prev_offset` is `sizeof(line_buffer) - 2` and we will still not enter the if-body. The fix is simple, just use `>=` instead of `==`.
* However, is escaping \ and " in the string needed at all? I don't think so and thus removed it, or do I miss something?
* Additionally moved the logging of "Console message received..." and the actual message to the place before calling the input handler. Because if the input handler also logs something the order is somewhat wrong, e.g. after sending "loadgame('something.sav')" in the console window:
    * "loadgame('something.sav') called"
    * "loading game..."
    * "Console message received..."
    * "loadgame('something.sav')"

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -492,7 +492,7 @@ static void AtomicSub(rmtS32 volatile* value, rmtS32 sub)
 }
 
 
-// Compiler write fences (windows implementation)
+// Compiler write fences
 static void WriteFence()
 {
 #if defined(RMT_PLATFORM_WINDOWS) && !defined(__MINGW32__)
@@ -5203,6 +5203,12 @@ RMT_API void _rmt_LogText(rmtPStr text)
     line_buffer[1] = 'O';
     line_buffer[2] = 'G';
     line_buffer[3] = 'M';
+    // Fill with spaces to enable viewing line_buffer without offset in a debugger
+    // (will be overwritten later by QueueLine/rmtMessageQueue_AllocMessage)
+    line_buffer[4] = ' ';
+    line_buffer[5] = ' ';
+    line_buffer[6] = ' ';
+    line_buffer[7] = ' ';
     start_offset = 8;
 
     // There might be newlines in the buffer, so split them into multiple network calls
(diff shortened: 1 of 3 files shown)


lib/Remotery.c
```diff
@@ -5205,51 +5206,32 @@ RMT_API void _rmt_LogText(rmtPStr text)
...
-                line_buffer[prev_offset++] = '\\';
-                line_buffer[prev_offset++] = '\"';
-                break;
+            offset = start_offset;
 
-            // Add the rest
-            default:
-                line_buffer[prev_offset++] = c;
-                break;
+            if (c == '\n')
```
dwilliamson: Can you leave a comment about what this is doing? The more readable code would be to check `c != '\n'` before adding to `line_buffer` below but this adds an extra branch per character.

dwilliamson:
Thank you for all these fixes. I've just added a single issue before merge.

lib/Remotery.c
```diff
@@ -5205,51 +5206,32 @@ RMT_API void _rmt_LogText(rmtPStr text)
...
     // There might be newlines in the buffer, so split them into multiple network calls
-    prev_offset = start_offset;
+    offset = start_offset;
     for (i = 0; text[i] != 0; i++)
     {
         char c = text[i];
 
         // Line wrap when too long or newline encountered
-        if (prev_offset == sizeof(line_buffer) - 3 || c == '\n')
+        if (offset == sizeof(line_buffer) || c == '\n')
```
dwilliamson: There used to be an extra -1 here so that `line_buffer` could be viewed in the debugger using the null termination above (`line_buffer = { 0 }`). Can you restore that?

I've also noticed that setting `start_offset = 8` leaves zeroes in-between so you can no longer view the string in a debugger without offsetting it but that's my bug!

tillrathmann: Ah, I see! Makes sense, I restored it and also added the spaces.

lib/Remotery.c
tillrathmann: Right, I wanted to avoid the extra branch per character that was not needed anymore after removing the \\ and \" escaping.
Added a comment now, though.

dwilliamson:
Great stuff. Thanks.

PR #150, opened 2019-02-01
Do not define rsize_t on FreeBSD

0mp:
It is already defined in string.h and leads to a compilation error.


dwilliamson:
Thanks, will get to this in the next few days. I'm leaving it open because I think `rsize_t` may need to be replaced with a name that doesn't clash with a C11 standard type.

PR #151, opened 2019-02-07
Fix/ported to compile and work as UWP App. 

tjachmann:
- Compilation fixes/required changes to run under UWP. 
- Successfully tested on Hololens


dwilliamson:
Nice one, thanks!

Issue #112, opened 2017-01-30
rmt_UnbindOpenGL blocks indefinitely


graphitemaster:
after creating Remotery and OpenGL context on the same thread, and issuing `rmt_BindOpenGL` on the same thread. Placing a call to `rmt_UnbindOpenGL` before shutting down Remotery blocks indefnintely in `Remotery_BlockingDeleteSampleTree` because of the following code:

```
            // Wait around until the Remotery server thread has sent all sample trees
            // of this type to the client
            while (sample_tree->allocator->nb_inuse > 1)
                msSleep(1);
```

The only way to get the application to close is to load up the Remotery web console in a webbrowser and wait for all the events to be sent to it. Is there a way to "jump the gun" and terminate the application even if there is samples pending?

graphitemaster:
Adding the following in `Remotery_FlushMessageQueue` fixes the problem

```
        // NOTE(daleweiler): added this
        if ( rmtMessageQueue_IsEmpty( rmt->mq_to_rmt_thread ) )
            break;
```

Then instead of breaking when message is NULL, continue (skip it) instead

graphitemaster:
Actually that does not fix it, it just takes a long time to release the queue

graphitemaster:
It's just confusing why it's not terminating immediately

dwilliamson:
So the Remotery thread owns the sample memory once your profiled thread hands it over. Once it's done its work it hands it back to your thread to release the queries on the thread that they were created. The queries don't *have* to follow this path, the code is just a lot simpler to write that way.

I'm presuming the main Remotery thread loop has exited as per your request and the server is looping in `Remotery_FlushMessageQueue`

Can you verify that?

graphitemaster:
Looks like it

graphitemaster:
It also looks like the text in the samples has disappeared in the web viewer as well, almost magically, kind of strange

dwilliamson:
I think I'd like to solve why it's not shutting down when no client is running, and then hopefully the rest will fall out from that.

Does it go into the same infinite loop with no viewer running?

graphitemaster:
It terminates when the viewer is running, it's when it's not running that is gets stuck in an infinite loop

dwilliamson:
That's confusing; when the main thread loop exits, there's no further interaction with the web client.

In `Remotery_FlushMessageQueue` can you verify that `rmt->mq_to_rmt_thread->write_pos` isn't advancing?

[...]

PR #155, opened 2019-02-15
Fix D3D11 sampling self length

ofoure:
Hi

Thanks for your works =)

Allow me to go straight to the point.
I had two issues: 
 - wrong self length for gpu sampling
 - infinite stall when unbinding D3D11

The fix for self length is simple and obvious but maybe it's not the case for the stall fix.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -6180,6 +6180,12 @@ static rmtBool GetD3D11SampleTimes(Sample* sample, rmtU64* out_first_timestamp, 
         sample->us_length = sample->us_end - sample->us_start;
     }
 
+    // Sum length on the parent to track un-sampled time in the parent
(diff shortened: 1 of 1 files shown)


dwilliamson:
Thanks for the PR!

I'd like to accept the self length fix but your second fix introduces a memory leak. The whole thing is a bit broken and you motivated me to get it fixed properly. See https://github.com/Celtoys/Remotery/issues/112 for more context.

I will try and get my fix in before the weekend closes. Meanwhile, could you trim the PR to just include the sample fix?

ofoure:
You'r more than welcome!

It's done, I've just reverted the fix for the stall regarding the memory leak.

dwilliamson:
The fix for shutdown has been submitted. cc @ofoure

PR #157, opened 2019-02-20
Add FreeBSD installation instructions

0mp:


dwilliamson:
Thanks for making this more available. Is the port in sync with the github repo, is there a release process or does it need to be constantly pulled?

0mp:
It's always fixed. At the moment it is fixed at the last commit from 2019-02-09 (38f3ed9). Ideally, I'd update the port when a new tag is released. Otherwise, I just update it from time to time as new commits are available upstream.

Here's the makefile that manages the port: https://svnweb.freebsd.org/ports/head/devel/remotery/Makefile?revision=493067&view=markup

A nice summary can be found here: https://www.freshports.org/devel/remotery/

PR #160, opened 2019-02-25
Avoid unaligned reads during hashing

### LINKED ISSUE #159: Violating pointer aliasing, possible UB
Stringified thread names get passed into MurmurHash3_x86_32 as const void *, which then assumes they're aligned by 4 bytes and casts them into (const rmtU32 *). This is undefined behavior, and with gcc7 -fsanitize=alignment build of my program which uses Remotery I get:

`external/Remotery/src/Remotery.c:2917:13: runtime error: load of misaligned address 0x0000042fe383 for type 'const rmtU32', which requires 4 byte alignment`

This is traced back to `rmt_BeginCPUSample(Wakeup, 0);` in `Remotery_ThreadMain`

yupinov:
Closes #159

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -2910,11 +2910,13 @@ static rmtU32 rotl32(rmtU32 x, rmtS8 r)
 }
 
 
-// Block read - if your platform needs to do endian-swapping or can only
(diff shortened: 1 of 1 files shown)


dwilliamson:
Thanks!

PR #162, opened 2019-03-08
Bugreport with quick workaround for race condition/read access violation. 

tjachmann:
The bug occurs in the following scenario, typically with many unique names being sampled (dynamic namings).

The profiled-thread calls **StringTable_Insert->Buffer_Write->Buffer_Grow** to insert a new name.
While at the same time the remotery-thread handles messages and calls **Remotery_ReceiveMessage->case GSMP**. Then the later accesses a buffer (the name string), which is currently being grown in the profiled thread and not in a defined state. Thus causing a memory access violation.

The proposed solution works practically all the time in all my scenarios. Though not guaranteed to work always. More changes or explicit synchronization would be needed therefore.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -1771,9 +1771,10 @@ typedef struct
     rmtU32 alloc_granularity;
 
     rmtU32 bytes_allocated;
-    rmtU32 bytes_used;
+    rmtU32 bytes_used;	
 
     rmtU8* data;
+	rmtU8* data_back;
 } Buffer;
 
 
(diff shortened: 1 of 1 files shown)


dwilliamson:
Oh my word, this was totally not on my radar! Problem is, the hash table isn't concurrency-safe either. I think the only way to safely handle this is by injecting an "add string table entry" message to the queue and then let the Remotery thread add it when it picks it up.

dwilliamson:
This was auto-closed when I changed the default branch from `master` to `main`.

However, any string table concurrency issues have been fixed as of the last few days.

PR #165, opened 2020-02-07
[Viewer] Add window resizing

jkunstwald:
rmt_resizing

- Added handle to bottom left of windows allowing click-and-drag resizing
- Cleaned up some CSS positioning logic
- Changed the window hide button from "O" to a unicode multiplication ` x`
- Added subtle hover highlights to buttons and the window hide button
- Changed window hide to only occur on left-click and mouseup instead of mousedown

Note that windows still snap back into their default size and position when the browser resizes.

### DIFF
--- a/vis/Code/SampleWindow.js
+++ b/vis/Code/SampleWindow.js
@@ -16,7 +16,7 @@ SampleWindow = (function()
         this.Visible = true;
 
         // Create a grid that's indexed by the unique sample ID
-        this.Grid = this.Window.AddControlNew(new WM.Grid(0, 0, 380, "calc( 100% - 17px )"));
+        this.Grid = this.Window.AddControlNew(new WM.Grid());
         var cell_data =
         {
             Name: "Samples",
--- a/vis/extern/BrowserLib/WindowManager/Code/Window.js
+++ b/vis/extern/BrowserLib/WindowManager/Code/Window.js
@@ -8,7 +8,7 @@ WM.Window = (function()
 		<div class='Window'>
 			<div class='WindowTitleBar'>
 				<div class='WindowTitleBarText notextsel' style='float:left'>Window Title Bar</div>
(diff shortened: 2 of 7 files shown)


vis/extern/BrowserLib/WindowManager/Code/Window.js
```diff
@@ -229,6 +278,8 @@ WM.Window = (function()
 
 	Window.prototype.SetSize = function(w, h)
 	{
+		w = (w < 80) ? 80 : w;
+		h = (h < 15) ? 15 : h;
 		this.Size = [ w, h ];
```
dwilliamson: Use `Math.max`?

vis/Code/SampleWindow.js
```diff
@@ -16,7 +16,7 @@ SampleWindow = (function()
         this.Visible = true;
 
         // Create a grid that's indexed by the unique sample ID
-        this.Grid = this.Window.AddControlNew(new WM.Grid(0, 0, 380, "calc( 100% - 17px )"));
+        this.Grid = this.Window.AddControlNew(new WM.Grid());
```
dwilliamson: I'm concerned this loses the ability for the grid control to resize with its container. The goal was to allow resizable columns but I never got around to that. Is it much of an issue? I will have to test!

vis/extern/BrowserLib/WindowManager/Code/Window.js
jkunstwald: Done!

vis/Code/SampleWindow.js
jkunstwald: I removed these arguments simply because they are no longer internally required in the grid. It looks like resizable columns aren't entirely straightforward (without dragging in libraries like jquery), but i don't think you'd need absolute sizes either way.

dwilliamson: Definitely not bringing in jquery but you're right about not needing absolute sizes.

PR #168, opened 2020-08-17
Add cmake support

intelligide:
Add a CMakeLists.txt for easier compilation.


dwilliamson:
Thanks for taking the time to make this PR but it's not something I'm looking to incorporate right now. I want to keep the repo clean of build configurations as the library caters to a huge number of platforms and use-cases. This means I really should delete the .bat sample builder as well...

PR #180, opened 2021-07-28
Ensure allocator pop doesn't try to pop from an empty list

dougbinks:
I've repeatedly hit a NULL dereference at line 2162 in Remotery.c on my WSL Linux builds (both Clang and GCC):
```
ObjectLink* next_link = old_link->next;
```

This is presumably because the function can be entered when `old_link` is non-NULL, but dereferenced when it is.

The solution is to ensure the pop can return NULL and the allocator can then allocate and directly return (or push/pop/retest in a loop but that seems not to be required).

Note that I may be seeing this when you're not due to having a machine with 36 hardware threads. However the logic above seems sound to me unless I'misunderstanding the overall picture.

I've tried to keep formatting to your conventions, but I may have missed some, my apologies.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -2153,12 +2153,15 @@ static ObjectLink* ObjectAllocator_Pop(ObjectAllocator* allocator)
     ObjectLink* link;
 
     assert(allocator != NULL);
-    assert(allocator->first_free != NULL);
 
     // CAS pop from the front of the list
     for (;;)
     {
         ObjectLink* old_link = (ObjectLink*)allocator->first_free;
+        if (old_link == NULL)
+        {
+            return NULL;
+        }
         ObjectLink* next_link = old_link->next;
         if (AtomicCompareAndSwapPointer((long* volatile*)&allocator->first_free, (long*)old_link, (long*)next_link) ==
             RMT_TRUE)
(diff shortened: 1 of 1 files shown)


dougbinks:
Note: After this PR Remotery works for a while on WSL Linux in my [enkiTSExamples](https://github.com/dougbinks/enkiTSExamples) but then fails at the `SampleTree_Pop` assert:
```
assert(sample != tree->root);
```

The Windows builds are fine. I don't think this is related to the PR, but the sample tree logic is too complex for me to parse in a limited amount of time.

dwilliamson:
Thanks, this is roughly in the area that I'm getting problem reports only on Linux.

I will clear some time to take a look at this in the morning.

erikogenvik:
> assert(sample != tree->root)



> Note: After this PR Remotery works for a while on WSL Linux in my [enkiTSExamples](https://github.com/dougbinks/enkiTSExamples) but then fails at the `SampleTree_Pop` assert:
> 
> ```
> assert(sample != tree->root);
> ```
> 
> The Windows builds are fine. I don't think this is related to the PR, but the sample tree logic is too complex for me to parse in a limited amount of time.

This issue should be fixed by https://github.com/Celtoys/Remotery/pull/181

dwilliamson:
Are any of these changes needed with #181 applied? I'm wary to change the behaviour as this hasn't been an issue for many years, only when I regressed with the thread sampling.

dougbinks:
The code without my change is incorrect if used on multiple threads - `ObjectAllocator_Pop` can be entered into by two or more threads with only one free `ObjectLink`, resulting in one thread popping successfully and the others dereferencing an `old_link` which is `NULL`.

If it's not used on multiple threads the atomics aren't needed.

dwilliamson:
OK, object allocators are not used from different threads; this looks like a hang over from a previous design.

Right now there is a sample tree per thread and sample trees are the only types to use object allocators.

So it looks like the Linux break inadvertently triggered the multithread path on the object allocator, which genuinely had the flaws you discovered.

I'm happy to accept this PR rather than go on a more destructive path to make the object allocator simpler. That can be left for another task as it would yield many other benefits.

dougbinks:
I've double checked and indeed this change isn't needed if rmtGetCurrentThreadId is corrected on POSIX, so in the long term removing the atomics should give some perf improvement - though an uncontended atomic is only a few times slower than a simple op.

dwilliamson:
The whole of `ObjectAllocator_Pop` will collapse into a simple single linked list pop, with no CAS loop, so no need to worry about whether branch prediction performance is an issue, or indeed having to attribute with likely/unlikely. It fits with the goal of pushing the hot path as close to a micro profiler as possible. Most of the other atomics can also be removed or used optionally in the build as they are only a debug tool, and don't actively contribute to the behaviour of the profiler.

PR #181, opened 2021-07-28
Thread id on nix

erikogenvik:
This fixes an issue where all threads on *NIX always got the same profiler.
Apart from random crazyness there would often be an assert of "assert(sample != tree->root);" in SampleTree_Pop after a while as different threads popped and pushed samples to the same profiler.

Note that I've tested the thread id code on Linux, but not on OSX or FreeBSD.

In addition I fixed an issue with the shader code that made the frontend stop working on some machines.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -1705,7 +1705,7 @@ static rmtThreadId rmtGetCurrentThreadId()
 #ifdef RMT_PLATFORM_WINDOWS
     return GetCurrentThreadId();
 #else
-    return pthread_self();
+    return (rmtThreadId)pthread_self();
 #endif
 }
 
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -126,6 +126,9 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
             #include <pthread_np.h>
         #else
             #include <sys/prctl.h>
+            //On Linux we need to do a syscall to get the thread id.
+            #include <sys/types.h>
+            #include <sys/syscall.h>
         #endif
     #endif
 
@@ -1705,7 +1708,12 @@ static rmtThreadId rmtGetCurrentThreadId()
 #ifdef RMT_PLATFORM_WINDOWS
     return GetCurrentThreadId();
 #else
-    return 0;
+#elif defined(RMT_PLATFORM_POSIX)
+    #if defined(__FreeBSD__) || defined(__OpenBSD__)
+    return pthread_getthreadid_np();
+    #else
+    return syscall(__NR_gettid);
+    #endif
 #endif
 }
 
@@ -4976,7 +4984,7 @@ static void ThreadProfiler_Destructor(ThreadProfiler* thread_profiler)
 static rmtError ThreadProfiler_Push(SampleTree* tree, rmtU32 name_hash, rmtU32 flags, Sample** sample)
 {
     rmtError error;
-    ModifySampleTree(tree, 
+    ModifySampleTree(tree,
         error = SampleTree_Push(tree, name_hash, flags, sample);
     );
     return error;
@@ -4993,7 +5001,7 @@ static rmtBool ThreadProfiler_Pop(ThreadProfiler* thread_profiler, rmtMessageQue
         Sample* root;
 
         // Disconnect all samples from the root and pack in the chosen message queue
-        ModifySampleTree(tree, 
+        ModifySampleTree(tree,
         root = tree->root;
         root->first_child = NULL;
         root->last_child = NULL;
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -126,9 +126,6 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
             #include <pthread_np.h>
         #else
             #include <sys/prctl.h>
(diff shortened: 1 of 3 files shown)


dwilliamson:
This is a brilliant catch, thanks. I'll look over this change in the next few hours.

lib/Remotery.c
```diff
@@ -1705,7 +1708,12 @@ static rmtThreadId rmtGetCurrentThreadId()
 #ifdef RMT_PLATFORM_WINDOWS
     return GetCurrentThreadId();
 #else
```
dougbinks, line 1710: Extra #else before #elif - needs removing

dougbinks:
I've just checked this out on top of my PR #180 and with both of these my test example now works fine, though on Unix I'm not seeing the processor timeline populate so there's likely still some missing functionality. As noted above there's a `#else` which needs removing.

dwilliamson:
I've been looking at `gettid`, `syscall` with `__NR_gettid`/`SYS_gettid`, and `pthread_threadid_np`. They seem to be a mess of incompatibilities across all the supported platforms. Could not this function be implemented in terms of `pthread_self` for all the POSIX platforms? It at least is already compiling and being used on the supported platforms. My thoughts are similar to this: http://elliotth.blogspot.com/2012/04/gettid-on-mac-os.html

lib/Remotery.c
```diff
@@ -1705,7 +1708,12 @@ static rmtThreadId rmtGetCurrentThreadId()
...
-    return 0;
+#elif defined(RMT_PLATFORM_POSIX)
```
dwilliamson, line 1711: I've been looking at gettid, syscall with __NR_gettid/SYS_gettid, and pthread_threadid_np. They seem to be a mess of incompatibilities across all the supported platforms. Could not this function be implemented in terms of pthread_self for all the POSIX platforms? It at least is already compiling and being used on the supported platforms. My thoughts are similar to this: http://elliotth.blogspot.com/2012/04/gettid-on-mac-os.html

vis/Code/PixelTimeRange.js
```diff
@@ -55,10 +55,13 @@ class PixelTimeRange
 
 	SetAsUniform(gl, program)
 	{
-		glSetUniform(gl, program, "inTimeRange.pxSpan", this.Span_px);
+	    //Uniforms that aren't used by the shaders are commented out.
```
dwilliamson, line 58: Can you remove `usSpan`, `pxSpan` and `usEnd` completely from the shader instead of leaving them commented out?

dwilliamson (changes requested):

dougbinks:
I've used `pthread_self()` before for this, and it works for Remotery too:
https://github.com/dougbinks/Remotery/commit/7ba46da401cc9a90dd0d0822b69d94acdd948327

vis/Code/PixelTimeRange.js
erikogenvik: Absolutely. I left them in because I wasn't sure if this was WIP code or not.

lib/Remotery.c
erikogenvik: My thinking here was that pthread_self returned a "pthread_t" which is an unsigned long int,  which might not fit into "rmtThreadId" which is a "unsigned int".
On the other hand, "syscall" also returns a "long int", so that might also be true there.
Perhaps better to use pthread_self on *NIX and then make rmtThreadId into a "rmtU64" just to be on the safe side?

dwilliamson: Yeah, I think making `rmtThreadId` a `rmtU64` on all platforms would be the safest option.

dwilliamson:
Almost perfect :)

erikogenvik:
> Almost perfect :)

Getting there. 😁
What did I miss?

dwilliamson (changes requested):

dwilliamson:
Gah, sorry, forgot to hit the submit button.

lib/Remotery.c
```diff
@@ -1705,7 +1705,7 @@ static rmtThreadId rmtGetCurrentThreadId()
 #ifdef RMT_PLATFORM_WINDOWS
     return GetCurrentThreadId();
 #else
-    return 0;
+    return pthread_self();
```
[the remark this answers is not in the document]
erikogenvik: Do you mean like
```
static rmtThreadId rmtGetCurrentThreadId()
{
#ifdef RMT_PLATFORM_WINDOWS
    return (rmtThreadId)GetCurrentThreadId();
#else
    return (rmtThreadId)pthread_self();
#endif
}
```
?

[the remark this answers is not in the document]
dwilliamson: Yes, there are some systems where `pthread_t` is a pointer type and not an integer type. For example I believe FreeBSD uses `typedef struct pthread* pthread_t`.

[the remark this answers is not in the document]
dwilliamson: Shouldn't need the cast on Windows version as that returns `DWORD`.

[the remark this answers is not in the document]
erikogenvik: Ah, I see. I'll fix that.

dwilliamson (approved):
Nice one, thanks.

PR #182, opened 2021-07-29
Fix ToolHelp32 usage in an Unicode environment Windows compile

noorus:
When compiling in an unicode environment on Windows, the MODULEENTRY32 struct uses widechar instead of char, and the strcpy in GetStartAddressModuleName doesn't work. This should fix that.


dwilliamson:
Thanks!

Issue #178, opened 2021-05-18
Assert failure on DestroyGlobalInstance when ThreadSampler is activated


septag:
I get a assert failure in `ObjectAllocator_Destructor` when ThreadSampler is activated: https://github.com/Celtoys/Remotery/blob/685fc6f2010c9b0235c89116b6fe3902690aa16c/lib/Remotery.c#L2121

Callstack:
![image](https://user-images.githubusercontent.com/5081921/118693366-89690400-b7fa-11eb-8251-8816d46b5d80.png)

Seems like some _ThreadProfiler_ trees are left hanging or lost and not deleted properly and eventually picked up by the destructor.

dwilliamson:
I have this issue as well, just no time to fix it right now.

Disabling the assert lets you continue and all memory is freed on shutdown (object allocations are from a pool).

It's definitely the thread sampling stalled sample detection. I believe it's even slowly leaking memory, which is the more serious problem.

PR #185, opened 2022-01-06
#178 Fix memory leaks when ThreadSampler is activated.

Archie3d:
This fixes the memory leaks in `SampleThreadsLoop`.
Two findings mainly:
- `stalling_sample_tree.root` does not get deleted when its children get queued;
- `processors` array is not deleted on return from the thread function.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -114,6 +114,7 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
         #undef max
         #include <tlhelp32.h>
         #include <winnt.h>
+        #include <timeapi.h>
         #include <processthreadsapi.h>
         #ifdef _XBOX_ONE
             #include "xmem.h"
@@ -1815,12 +1816,12 @@ DWORD_PTR GetThreadStartAddress(rmtThreadHandle thread_handle)
 
     return 0;
 }
- 
+
 const char* GetStartAddressModuleName(DWORD_PTR start_address)
 {
     BOOL success;
     MODULEENTRY32 module_entry;
- 
+
     // Snapshot the modules
     HANDLE handle = CreateToolhelp32Snapshot(TH32CS_SNAPMODULE, 0);
     if (handle == INVALID_HANDLE_VALUE)
@@ -1853,7 +1854,7 @@ const char* GetStartAddressModuleName(DWORD_PTR start_address)
 
         success = Module32Next(handle, &module_entry);
     }
- 
+
     CloseHandle(handle);
 
     return NULL;
@@ -4872,7 +4873,7 @@ typedef struct ThreadProfiler
     rmtU64 registerBackup0;                                                                         // 0
     rmtU64 registerBackup1;                                                                         // 8
     rmtU64 registerBackup2;                                                                         // 16
-    
+
     // Used to schedule callbacks taking into account some threads may be sleeping
     rmtS32 nbSamplesWithoutCallback;                                                                // 24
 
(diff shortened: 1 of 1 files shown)


dwilliamson:
Thanks for this PR, most helpful. Please give me a day to review the changes.

lib/Remotery.c
```diff
@@ -5769,6 +5772,9 @@ static rmtError SampleThreadsLoop(rmtThread* rmt_thread)
                 Sample* sample = stalling_sample_tree.root->first_child;
                 assert(sample != NULL);
                 QueueSampleTree(thread_profilers->mqToRmtThread, sample, stalling_sample_tree.allocator, thread_profiler->threadName, thread_profiler, RMT_TRUE);
+
+                assert(stalling_sample_tree.allocator != NULL);
+                ObjectAllocator_Free(stalling_sample_tree.allocator, stalling_sample_tree.root);
```
dwilliamson, line 5777: Can I verify that your understanding of this change is that:

1. `stalling_sample_tree.root->first_child` has been sent to the main Remotery thread via `QueueSampleTree`.
2. This will get released later when the Remotery thread has finished with it.
3. This leaves `stalling_sample_tree.root` allocated here and nobody frees it, leading to memory leak.

Your change here adds that required free call.

lib/Remotery.c
```diff
@@ -5757,7 +5757,10 @@ static rmtError SampleThreadsLoop(rmtThread* rmt_thread)
             // While the thread is suspended take the chance to check for samples trees that may never complete
             // Because SuspendThread on Windows is an async request, this needs to be placed at a point where the request completes
             // Calling GetThreadContext will ensure the request is completed so this stall check is placed after that
-            CheckForStallingSamples(&stalling_sample_tree, thread_profiler, sample_time_us);
+            if (RMT_ERROR_NONE != CheckForStallingSamples(&stalling_sample_tree, thread_profiler, sample_time_us))
+            {
+                SampleTree_Destructor(&stalling_sample_tree);
```
dwilliamson, line 5762: I don't believe this is correct:

1. `stalling_sample_tree` is allocated via `SampleTree_Copy`.
2. This *aliases* the `allocator` copied from the source tree.
3. As such; `stalling_sample_tree` does not own its `allocator` pointer.
4. `SampleTree_Destructor` explicitly calls `Delete` on the `allocator` pointer, which it doesn't own.

I think the best course of action here is to call `FreeSamples` instead of `SampleTree_Destructor`.

lib/Remotery.c
Archie3d: This is correct, the `root` sample is leaking here.

lib/Remotery.c
Archie3d: I see. This actually never happens (in my tests) but since `CheckForStallingSamples` may return with an error I beleive releasing the samples tree is the right thing to do here.

Should the upper bit releasing the `root` be using the `FreeSamples` as well (instead of the `ObjectAllocator_Free` directly)?

dwilliamson: `FreeSamples` is recursive: it will flatten an entire tree of samples and then walk that list and free them each.

Given that you're only releasing the root, your call to `ObjectAllocator_Free` is already achieving its goal.

However, calling `FreeSamples` there is probably a good idea also as the API can be cleaned up at a later date.

lib/Remotery.c
dwilliamson: Great, could you leave a comment to that effect? The Sample Tree API needs cleaning up and this will help a later refactor.

Archie3d: Sure. Left the comment in there.

lib/Remotery.c
Archie3d: Great. I change it to use `FreeSamples` in there as well then.

dwilliamson (approved):

Issue #190, opened 2022-04-06
Api for iterating the profiler info


JCash:
In our current profile library, we have support for iterating the samples and counters.
This allows us to display the info quickly at runtime. E.g.:

![image](https://user-images.githubusercontent.com/1349334/161962141-070d8668-86d8-4750-8927-0787c0db7cc4.png)

Is this a feature you've considered before, and if so, what were/are your current thoughts around such an api in remotery?

Regards,
Mathias

dwilliamson:
Right now I'm not convinced this is a good way forward for the API as it would pull a lot of internal structures into the public API, such as the Sample Tree. Any attempt to make some separation between public and private versions of these would complicate it even more.

A lot of people wrap remotery in their own macros so that they can swap profilers in and out. So some might do something like:

```cpp
#define MySample rmt_ScopedCPUSample
```

In this case I would suggest using `MySample` to push your samples into your own tree structure as well.

JCash:
Yeah, for sure, keeping a stable api is the trickier part of it all.
However, seeing as the update rate to the library is very stable it might be worth considering.

Definitely, keeping our current profiler is of course an option.
It just feels strange to keep two profilers around.
In any case thanks for the heads up!

dwilliamson:
Oh, so you want to inspect the values recorded by Remotery as well? Hmmm... there is a precedent for this with people actually piggybacking the network stream to receive all the updates.

JCash:
Ah, yes.
So our current implementation is basically the same as Remotery, so we have all the defines and hooks as expected.
But our lib lacks a few features, and I'd much rather use something commonly used (more modern UI, better license etc).

I'm not opposed to using a fork of Remotery either, where I can traverse the internal structures, at least to get a feel for it (I am currently in the due diligence phase). In reality, I don't expect the internals changing too much too often, and if they do, we'll have to pay that time it takes.
As long as I know beforehand, I can at least make informed choices :)

JCash:
Yeah, I wouldn't want to expose the internals of the `Sample`either, but more likely add it as an opaque pointer that can be used to query the info needed. It would then probably require the enum `SampleType` to be available as well.

The `SAMPLE_NAME_LEN` might as well be a customizable define in the header I think? That would also make it known to the user.

I was thinking something along these lines:
```


// Iterating the tree
// Allows the caller to keep track of the hierarchy
// Based on Per Vognsen's gist:
[...]

PR #191, opened 2022-04-07
Added sample interator api

JCash:
Fixes https://github.com/Celtoys/Remotery/issues/190

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -6107,7 +6107,7 @@ static rmtError Remotery_SendSampleTreeMessage(Remotery* rmt, Message* message)
 
     if (g_Settings.sampletree_handler != NULL)
     {
-        g_Settings.sampletree_handler(rmt, g_Settings.sampletree_context, sample_tree);
+        g_Settings.sampletree_handler(g_Settings.sampletree_context, sample_tree);
     }
 
     // Release sample tree samples back to their allocator
@@ -8843,12 +8843,12 @@ RMT_API rmtBool _rmt_IterateNext(rmtSampleIterator* iter)
 }
 
 // Sample tree accessors
-RMT_API const char* _rmt_SampleTreeGetThreadName(rmtMsgSampleTree* sample_tree)
+RMT_API const char* _rmt_SampleTreeGetThreadName(rmtSampleTree* sample_tree)
 {
     return sample_tree->threadName;
 }
 
-RMT_API rmtSample* _rmt_SampleTreeGetRootSample(rmtMsgSampleTree* sample_tree)
+RMT_API rmtSample* _rmt_SampleTreeGetRootSample(rmtSampleTree* sample_tree)
 {
     return sample_tree->rootSample;
 }
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -4309,22 +4309,12 @@ static void Server_Update(Server* server)
 
 #define SAMPLE_NAME_LEN 128
 
-typedef enum SampleType
-{
-    SampleType_CPU,
-    SampleType_CUDA,
-    SampleType_D3D11,
-    SampleType_OpenGL,
-    SampleType_Metal,
-    SampleType_Count,
-} SampleType;
-
 typedef struct Sample
 {
     // Inherit so that samples can be quickly allocated
     ObjectLink Link;
 
-    enum SampleType type;
+    enum rmtSampleType type;
 
     // Hash generated from sample name
     rmtU32 name_hash;
@@ -4368,7 +4358,7 @@ static rmtError Sample_Constructor(Sample* sample)
 
     ObjectLink_Constructor((ObjectLink*)sample);
 
-    sample->type = SampleType_CPU;
+    sample->type = RMT_SampleType_CPU;
     sample->name_hash = 0;
     sample->unique_id = 0;
     sample->unique_id_html_colour[0] = '#';
@@ -4885,7 +4875,7 @@ typedef struct ThreadProfiler
     rmtU32 threadNameHash;
 
     // Store a unique sample tree for each type
-    SampleTree* sampleTrees[SampleType_Count];
+    SampleTree* sampleTrees[RMT_SampleType_Count];
 
 #if RMT_USE_D3D11
     D3D11* d3d11;
@@ -4947,7 +4937,7 @@ static rmtError ThreadProfiler_Constructor(rmtMessageQueue* mq_to_rmt, ThreadPro
     QueueThreadName(mq_to_rmt, thread_profiler->threadName, thread_profiler);
 
     // Create the CPU sample tree only. The rest are created on-demand as they need extra context to function correctly.
-    New_3(SampleTree, thread_profiler->sampleTrees[SampleType_CPU], sizeof(Sample), (ObjConstructor)Sample_Constructor,
+    New_3(SampleTree, thread_profiler->sampleTrees[RMT_SampleType_CPU], sizeof(Sample), (ObjConstructor)Sample_Constructor,
           (ObjDestructor)Sample_Destructor);
     if (error != RMT_ERROR_NONE)
     {
@@ -4973,7 +4963,7 @@ static void ThreadProfiler_Destructor(ThreadProfiler* thread_profiler)
     Delete(D3D11, thread_profiler->d3d11);
 #endif
 
-    for (index = 0; index < SampleType_Count; index++)
+    for (index = 0; index < RMT_SampleType_Count; index++)
     {
         Delete(SampleTree, thread_profiler->sampleTrees[index]);
     }
@@ -5536,7 +5526,7 @@ static rmtError CheckForStallingSamples(SampleTree* stalling_sample_tree, Thread
     stalling_sample_tree->allocator = NULL;
 
     // Skip the stall check if the tree is being modified
-    sample_tree = thread_profiler->sampleTrees[SampleType_CPU];
+    sample_tree = thread_profiler->sampleTrees[RMT_SampleType_CPU];
     if (LoadAcquire(&sample_tree->treeBeingModified) != 0)
     {
         return RMT_ERROR_NONE;
@@ -6025,19 +6015,19 @@ static rmtError bin_SampleTree(Buffer* buffer, Msg_SampleTree* msg)
     // Add any sample types as a thread name post-fix to ensure they get their own viewer
     thread_name[0] = 0;
     strncat_s(thread_name, sizeof(thread_name), msg->threadName, strnlen_s(msg->threadName, 255));
-    if (root_sample->type == SampleType_CUDA)
+    if (root_sample->type == RMT_SampleType_CUDA)
     {
         strncat_s(thread_name, sizeof(thread_name), " (CUDA)", 7);
     }
-    if (root_sample->type == SampleType_D3D11)
+    if (root_sample->type == RMT_SampleType_D3D11)
     {
         strncat_s(thread_name, sizeof(thread_name), " (D3D11)", 8);
     }
-    if (root_sample->type == SampleType_OpenGL)
+    if (root_sample->type == RMT_SampleType_OpenGL)
     {
         strncat_s(thread_name, sizeof(thread_name), " (OpenGL)", 9);
     }
-    if (root_sample->type == SampleType_Metal)
+    if (root_sample->type == RMT_SampleType_Metal)
     {
         strncat_s(thread_name, sizeof(thread_name), " (Metal)", 8);
     }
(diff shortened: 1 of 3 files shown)


lib/Remotery.c
```diff
@@ -4309,22 +4309,12 @@ static void Server_Update(Server* server)
 
 #define SAMPLE_NAME_LEN 128
 
-typedef enum SampleType
```
JCash, line 4312: Moved to the header, and renamed to `rmtSampleType`

lib/Remotery.c
```diff
@@ -6115,6 +6105,11 @@ static rmtError Remotery_SendSampleTreeMessage(Remotery* rmt, Message* message)
     error = bin_SampleTree(bin_buf, sample_tree);
     rmt_EndCPUSample();
 
+    if (g_Settings.sampletree_handler != NULL)
+    {
+        g_Settings.sampletree_handler(rmt, g_Settings.sampletree_context, sample_tree);
+    }
```
JCash, line 6111: I had to put it before the `FreeSamples()` as the sample pointers would be invalid after that call.

lib/Remotery.c
```diff
@@ -8820,4 +8817,92 @@ RMT_API void _rmt_EndMetalSample(void)
...
+        iter->initial = 0;
+    }
+    else
+    {
+        if (iter->sample != NULL)
+            iter->sample = iter->sample->next_sibling;
+    }
+
+    return iter->sample != NULL ? RMT_TRUE : RMT_FALSE;
+}
```
JCash, line 8843: Following the pattern from this gist:
https://gist.github.com/pervognsen/744066869da509a15c4e08fc9408207e

lib/Remotery.c
```diff
@@ -8820,4 +8817,92 @@ RMT_API void _rmt_EndMetalSample(void)
...
+// Sample tree accessors
+RMT_API const char* _rmt_SampleTreeGetThreadName(rmtMsgSampleTree* sample_tree)
+{
+    return sample_tree->threadName;
+}
+
+RMT_API rmtSample* _rmt_SampleTreeGetRootSample(rmtMsgSampleTree* sample_tree)
+{
+    return sample_tree->rootSample;
+}
```
JCash, line 8854: At first, I passed the `threadName` explicitly to the callback function, but later thought that passing the opaque message pointer fits the api a bit better:
* Easier to add more accessor functions later
* Matches the Sample accessor api used in the callback api

lib/Remotery.c
```diff
@@ -8820,4 +8817,92 @@ RMT_API void _rmt_EndMetalSample(void)
...
+// Sample accessors
+RMT_API const char* _rmt_SampleGetName(Remotery* rmt, rmtSample* sample)
+{
+    const char* name = StringTable_Find(rmt->string_table, sample->name_hash);
+    if (name == NULL)
+    {
+        return "null";
+    }
+    return name;
+}
```
JCash, line 8865: Needs the Remotery pointer, so I explicitly pass that in the callback.

lib/Remotery.c
```diff
@@ -8820,4 +8817,92 @@ RMT_API void _rmt_EndMetalSample(void)
...
+
+RMT_API rmtU64 _rmt_SampleGetTime(rmtSample* sample)
+{
+    return sample->us_length;
+}
+
+RMT_API rmtU64 _rmt_SampleGetSelfTime(rmtSample* sample)
+{
+    return (rmtU64)maxS64(sample->us_length - sample->us_sampled_length, 0);
```
JCash, line 8884: I think this is correct, as it's the same as is sent over the socket.

lib/Remotery.h
```diff
@@ -311,6 +316,26 @@ typedef enum rmtSampleFlags
...
 
+typedef enum rmtSampleType
+{
+    RMT_SampleType_CPU,
+    RMT_SampleType_CUDA,
+    RMT_SampleType_D3D11,
+    RMT_SampleType_OpenGL,
+    RMT_SampleType_Metal,
+    RMT_SampleType_Count,
+} rmtSampleType;
```
JCash, line 327: Given the other public interfaces in the header, I prefixed it with RMT_.

Might want to be upper case, to match style of `rmtError` enum?

lib/Remotery.h
```diff
@@ -363,6 +388,7 @@ typedef void* (*rmtMallocPtr)(void* mm_context, rmtU32 size);
 typedef void* (*rmtReallocPtr)(void* mm_context, void* ptr, rmtU32 size);
 typedef void (*rmtFreePtr)(void* mm_context, void* ptr);
 typedef void (*rmtInputHandlerPtr)(const char* text, void* context);
+typedef void (*rmtSampleTreeHandlerPtr)(Remotery* rmt, void* cbk_context, rmtMsgSampleTree* sample_tree);
```
JCash, line 391: The callback type and its arguments.

lib/Remotery.h
```diff
@@ -661,6 +691,23 @@ RMT_API void _rmt_BeginMetalSample(rmtPStr name, rmtU32* hash_cache);
...
+RMT_API rmtSample*          _rmt_SampleTreeGetRootSample(rmtMsgSampleTree* sample_tree);
+
+// Sample accessors
+RMT_API const char*         _rmt_SampleGetName(Remotery* rmt, rmtSample* sample);
+RMT_API rmtU32              _rmt_SampleGetNameHash(rmtSample* sample);
+RMT_API rmtU32              _rmt_SampleGetCallCount(rmtSample* sample);
+RMT_API rmtU64              _rmt_SampleGetTime(rmtSample* sample);
+RMT_API rmtU64              _rmt_SampleGetSelfTime(rmtSample* sample);
+RMT_API void                _rmt_SampleGetColour(rmtSample* sample, rmtU8* r, rmtU8* g, rmtU8* b);
+RMT_API rmtSampleType       _rmt_SampleGetType(rmtSample* sample);
```
JCash, line 709: Currently, they all are "private", with the "_" prefix.

I guess they should also be wrapped with the "RMT_OPTIONAL" define.
Should it also be contained withing a new tag, e.g. `RMT_ENABLED_ITERATORS` ? Might be a bit overkill?

sample/dump.c
```diff
@@ -0,0 +1,113 @@
...
+
+void dumpTree(Remotery* rmt, void* ctx, rmtMsgSampleTree* sample_tree)
+{
+    rmtSample* root = _rmt_SampleTreeGetRootSample(sample_tree);
+    const char* thread_name = _rmt_SampleTreeGetThreadName(sample_tree);
+
+    printf("// ********************   DUMP TREE: %s   ************************\n", thread_name);
+
+    printTree(rmt, root, 0);
+}
```
JCash, line 75: Example callback

sample/dump.c
```diff
@@ -0,0 +1,113 @@
...
+
+void printTree(Remotery* rmt, rmtSample* sample, int indent)
+{
+    printSample(rmt, sample, indent);
+
+    rmtSampleIterator iter = _rmt_IterateChildren(sample);
+    while (_rmt_IterateNext(&iter)) {
+        printTree(rmt, iter.sample, indent+1);
+    }
+}
```
JCash, line 65: Example of iterator

lib/Remotery.h
```diff
@@ -661,6 +691,23 @@ RMT_API void _rmt_BeginMetalSample(rmtPStr name, rmtU32* hash_cache);
 RMT_API void _rmt_EndMetalSample(void);
 #endif
 
+// Iterator
```
dwilliamson, line 694: Does this API compile and link when `RMT_ENABLED=0`? Prefix-underscores shouldn't be used here.

As an example:

```cpp
RMT_API void _rmt_LogText(rmtPStr text);
```

This is the function signature that is part of the Private interface and this is the macro which gets compiled-out when `RMT_ENABLED=0`:

```cpp
#define rmt_LogText(text)                                                           \
    RMT_OPTIONAL(RMT_ENABLED, _rmt_LogText(text))
```

dwilliamson:
Really great stuff.

I started reviewing this before you left your comments so I'll address them now.

lib/Remotery.c
dwilliamson: Hmmm... yes... that's a good counter argument. I'll edit my comment.

lib/Remotery.c
```diff
@@ -6115,6 +6105,11 @@ static rmtError Remotery_SendSampleTreeMessage(Remotery* rmt, Message* message)
     error = bin_SampleTree(bin_buf, sample_tree);
     rmt_EndCPUSample();
 
+    if (g_Settings.sampletree_handler != NULL)
+    {
+        g_Settings.sampletree_handler(rmt, g_Settings.sampletree_context, sample_tree);
```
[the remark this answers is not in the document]
dwilliamson: Leave as is but could you rename to `rmtSampleTree`?

lib/Remotery.c
dwilliamson: You should use `g_Remotery` in the implementation instead.

lib/Remotery.h
dwilliamson: This is fine. I wanted to remove all error-messages-as-enums, anyway and replace them with 4 or 5 broader error enums that give you optional message retrieval.

lib/Remotery.h
dwilliamson: Don't worry about adding a `RMT_ENABLED_ITERATORS` - just `RMT_ENABLED` will do.

lib/Remotery.c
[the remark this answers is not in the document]
JCash: Sure. That's actually exactly what I had before :)

lib/Remotery.c
JCash: I was thinking that too, but I was perhaps trying to remove the need to use a global variable.

The current setup in the PR was to address the initialization problem:


The problem with the user context and the remotery pointer is a catch 22:
```
// Setting up settings before creating the context
Remotery *rmt;
rmtSettings* settings = rmt_Settings();
settings->sampletree_handler = dumpTree;
settings->sampletree_context = rmt; // Not valid yet
```

An alternative, albeit not very clean code flow either:
```
// Setup some setting variables before creating the Remotery instance...
Remotery *rmt;
rmtSettings* settings = rmt_Settings();
settings->port = port;

// ... and set some settings after
settings->sampletree_handler = dumpTree;
settings->sampletree_context = rmt; // Not valid yet
```

But, I have no examples of other users use case, and they can for sure create their own custom contexts etc.

So I'll go ahead with removing the Remotery pointer from the callback function.

lib/Remotery.h
JCash: I have now added this, but I'm wondering if the use case is a bit different.
The client code will fail to compile on code like this if we simply omit the info:

```
error: expected expression
    const char* name = rmt_SampleGetName(ctx->rmt, sample);
```

I see two options:
1. Each such interface needs to produce "valid" values, which is counter intuitive when using the `RMT_OPTIONAL + RMT_ENABLED`
2. Letting the client control their code using `#ifdef RMT_ENABLED` explicitly around this particular piece of code.

dwilliamson: Yes, there's going to be a lot of errors if you implement some of them in terms of `RMT_OPTIONAL` as they need to return a value even when `RMT_ENABLED=0`.

One option is to return nothing and specify the output as an address parameter to the function call but that would then prevent you from initialising at the declaration point in C when multiple calls are involved.

Have you looked at using `RMT_OPTIONAL_RET`?

lib/Remotery.c
dwilliamson: Most of the internal code uses the global variable `g_Remotery`. The API explicitly avoids requesting it from the user. It even has a DLL API to cater for this where the use calls `rmt_SetGlobalInstance`.

JCash: Ah, now I finally see what you mean. Fixing...

lib/Remotery.h
```diff
@@ -358,11 +383,49 @@ typedef enum rmtSampleFlags
     RMT_OPTIONAL(RMT_ENABLED, _rmt_EndCPUSample())
 
 
+#define rmt_IterateChildren(iter, sample)                                           \
+    RMT_OPTIONAL(RMT_ENABLED, _rmt_IterateChildren(iter, sample))
```
JCash, line 387: Tweaked the api from
```
rmtSampleIterator iter = rmt_IterateChildren(sample);
```
to
```
rmtSampleIterator iter;
rmt_IterateChildren(&iter, sample);
```
in order to both make it more C89 compatible, and also make it work with RMT_OPTIONAL better.

lib/Remotery.h
JCash: I updated the api now to use the RMT_OPTIONAL+RMT_OPTIONAL_RET and the dump.c example builds and runs in `-DRMT_ENABLED=0` as well.

dwilliamson:
I love this... going to get some tea and make sure I understand the implications of adding this before merge.

Thanks so much.

dwilliamson (approved):

PR #194, opened 2022-04-14
Add first proof of concept for rmt_StatI32

JCash:
The iterator example now outputs stats as well:
```
~/code/Remotery $ ./test
// ********************   DUMP TREE: Thread0   ************************
SAMPLE: delay 1  time: 12  self: 12 type: 0  color: 0xddf9f8
  SAMPLE: recursive 6  time: 0  self: 0 type: 0  color: 0x6e745f
  SAMPLE: aggregate 3  time: 0  self: 0 type: 0  color: 0xcfd6ff
    STAT: MyCounter type: RMT_StatType_I32 value: 3  desc: test
```

Also, here's a first screenshot of a statistic in action:
Screenshot 2022-04-14 at 16 06 15

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -4316,6 +4316,7 @@ typedef struct StatInfo
     enum rmtStatType type;
     union {
         rmtI32 ivalue;
+        rmtU32 uvalue;
         rmtF32 fvalue;
     } value;
     rmtU32 flags;
@@ -4476,12 +4477,20 @@ static void Sample_CopyState(Sample* dst_sample, const Sample* src_sample)
 
 static rmtError bin_SampleArray(Buffer* buffer, Sample* parent_sample);
 
+static const char* gStatTypeStrings[] = {
+    "I32",
+    "F32",
+    "TXT",
+    "NUL",
+};
+
 static rmtError bin_Sample(Buffer* buffer, Sample* sample)
 {
     rmtError error;
 
     assert(sample != NULL);
 
+    // For decoding, See Remotery.js : DecodeSample
     BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->name_hash));
     BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->unique_id));
     BIN_ERROR_CHECK(Buffer_Write(buffer, sample->unique_id_html_colour, 7));
@@ -4490,6 +4499,9 @@ static rmtError bin_Sample(Buffer* buffer, Sample* sample)
     BIN_ERROR_CHECK(Buffer_WriteU64(buffer, maxS64(sample->us_length - sample->us_sampled_length, 0)));
     BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->call_count));
     BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->max_recurse_depth));
+    BIN_ERROR_CHECK(Buffer_Write(buffer, gStatTypeStrings[sample->stat_info.type], 3));
+    BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->stat_info.value.uvalue));
+    BIN_ERROR_CHECK(Buffer_WriteU32(buffer, sample->stat_info.desc));
     BIN_ERROR_CHECK(bin_SampleArray(buffer, sample));
 
     return RMT_ERROR_NONE;
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -32,6 +32,7 @@
     @DYNBUF:        Dynamic Buffer
     @HASHTABLE:     Integer pair hash map for inserts/finds. No removes for added simplicity.
     @STRINGTABLE:	Map from string hash to string offset in local buffer
+    @ATOMICTABLE:   Atomic protected hash table
     @SOCKETS:       Sockets TCP/IP Wrapper
     @SHA1:          SHA-1 Cryptographic Hash Function
     @BASE64:        Base-64 encoder
@@ -2677,6 +2678,76 @@ static void StringTable_Insert(StringTable* table, rmtU32 name_hash, rmtPStr nam
 /*
 ------------------------------------------------------------------------------------------------------------------------
 ------------------------------------------------------------------------------------------------------------------------
+   @ATOMICTABLE: Map from string hash to string offset in local buffer
+------------------------------------------------------------------------------------------------------------------------
+------------------------------------------------------------------------------------------------------------------------
+*/
+
+typedef struct
+{
+    // Map from text hash to text location in the buffer
+    rmtHashTable* map;
+
+    rmtU32 lock;
+} AtomicTable;
+
+static void AtomicSpinLock(rmtU32 volatile* val)
+{
+    while (AtomicCompareAndSwap(val, 0, 1) == RMT_FALSE) {
+    }
+}
+
+static void AtomicSpinUnlock(rmtU32 volatile* val)
+{
+    *val = 0;
+}
+
+static rmtError AtomicTable_Constructor(AtomicTable* table, rmtU32 max_nb_slots)
+{
+    rmtError error;
+
+    // Default initialise
+    assert(table != NULL);
+    table->map = NULL;
+    table->lock = 0;
+
+    New_1(rmtHashTable, table->map, max_nb_slots);
+    if (error != RMT_ERROR_NONE)
+        return error;
+
+    return RMT_ERROR_NONE;
+}
+
+static void AtomicTable_Destructor(AtomicTable* table)
+{
+    assert(table != NULL);
+    AtomicSpinLock((rmtU32 volatile*)&table->lock);
+    Delete(rmtHashTable, table->map);
+    AtomicSpinUnlock((rmtU32 volatile*)&table->lock);
+}
+
+static rmtU64 AtomicTable_Find(AtomicTable* table, rmtU32 name_hash)
+{
+    AtomicSpinLock((rmtU32 volatile*)&table->lock);
+    rmtU64 match = rmtHashTable_Find(table->map, name_hash);
+    AtomicSpinUnlock((rmtU32 volatile*)&table->lock);
+    return match;
+}
+
+static void AtomicTable_Insert(AtomicTable* table, rmtU32 hash, rmtU64 value)
+{
+    AtomicSpinLock((rmtU32 volatile*)&table->lock);
+    rmtU64 text_offset = rmtHashTable_Find(table->map, hash);
+    if (text_offset == RMT_NOT_FOUND)
+    {
+        rmtHashTable_Insert(table->map, hash, value);
+    }
+    AtomicSpinUnlock((rmtU32 volatile*)&table->lock);
+}
+
+/*
+------------------------------------------------------------------------------------------------------------------------
+------------------------------------------------------------------------------------------------------------------------
    @SOCKETS: Sockets TCP/IP Wrapper
 ------------------------------------------------------------------------------------------------------------------------
 ------------------------------------------------------------------------------------------------------------------------
@@ -4311,16 +4382,21 @@ static void Server_Update(Server* server)
 
 #define SAMPLE_NAME_LEN 128
 
+typedef union {
+    rmtS32 ivalue;
+    rmtU32 uvalue;
+    rmtF32 fvalue;
+} StatValue;
+
 typedef struct StatInfo
 {
-    enum rmtStatType type;
-    union {
-        rmtI32 ivalue;
-        rmtU32 uvalue;
-        rmtF32 fvalue;
-    } value;
-    rmtU32 flags;
-    rmtU32 desc; // The hash of the description
+    // Points to the default value
+    struct Sample*          declaration;
+    StatValue               value;
+    rmtU32                  group_hash;
+    rmtU32                  flags;
+    rmtU32                  desc; // The hash of the description
+    enum rmtStatType        type;
 } StatInfo;
 
 typedef struct Sample
@@ -4368,12 +4444,14 @@ typedef struct Sample
 
 } Sample;
 
+
 static rmtError Sample_Constructor(Sample* sample)
 {
     assert(sample != NULL);
 
     ObjectLink_Constructor((ObjectLink*)sample);
 
+// TODO: Could we call a Sample_Clear() here?
     sample->type = RMT_SampleType_CPU;
     sample->name_hash = 0;
     sample->unique_id = 0;
@@ -4394,6 +4472,10 @@ static rmtError Sample_Constructor(Sample* sample)
     sample->max_recurse_depth = 0;
     sample->stat_info.type = RMT_StatType_Count;
     sample->stat_info.desc = 0;
+    sample->stat_info.flags = 0;
+    sample->stat_info.group_hash = 0;
+    sample->stat_info.value.ivalue = 0;
+    sample->stat_info.declaration = NULL;
 
     return RMT_ERROR_NONE;
 }
@@ -4403,8 +4485,27 @@ static void Sample_Destructor(Sample* sample)
     RMT_UNREFERENCED_PARAMETER(sample);
 }
 
+static rmtError Stat_Constructor(Sample* sample)
+{
+    assert(sample != NULL);
+
+    // Chain to sample constructor
+    Sample_Constructor(sample);
+    sample->type = RMT_SampleType_Stat;
+    sample->stat_info.type = RMT_StatType_Count;
+
+    return RMT_ERROR_NONE;
+}
+
+static void Stat_Destructor(Sample* sample)
+{
+    Sample_Destructor(sample);
+}
+
 static void Sample_Prepare(Sample* sample, rmtU32 name_hash, Sample* parent)
 {
+// TODO: Could we call a Sample_Clear() here, followed by setting the actual values?
+// I argue that the list of settings is getting long, and might be better suited to be cleared in one place
     sample->name_hash = name_hash;
     sample->unique_id = 0;
     sample->parent = parent;
@@ -4421,6 +4522,10 @@ static void Sample_Prepare(Sample* sample, rmtU32 name_hash, Sample* parent)
     sample->max_recurse_depth = 0;
     sample->stat_info.type = RMT_StatType_Count;
     sample->stat_info.desc = 0;
+    sample->stat_info.flags = 0;
+    sample->stat_info.group_hash = 0;
+    sample->stat_info.value.ivalue = 0;
+    sample->stat_info.declaration = NULL;
 }
 
 static void Sample_Close(Sample* sample, rmtU64 us_end)
@@ -4460,6 +4565,7 @@ static void Sample_CopyState(Sample* dst_sample, const Sample* src_sample)
     dst_sample->call_count = src_sample->call_count;
     dst_sample->recurse_depth = src_sample->recurse_depth;
     dst_sample->max_recurse_depth = src_sample->max_recurse_depth;
+    dst_sample->stat_info = src_sample->stat_info;
 
     // Prepare empty tree links
     dst_sample->parent = NULL;
@@ -4478,9 +4584,10 @@ static void Sample_CopyState(Sample* dst_sample, const Sample* src_sample)
 static rmtError bin_SampleArray(Buffer* buffer, Sample* parent_sample);
 
 static const char* gStatTypeStrings[] = {
-    "I32",
+    "S32",
     "F32",
     "TXT",
+    "GRP",
     "NUL",
 };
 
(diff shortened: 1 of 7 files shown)


lib/Remotery.c
```diff
@@ -4309,6 +4311,17 @@ static void Server_Update(Server* server)
...
 
+typedef struct StatInfo
+{
+    enum rmtStatType type;
+    union {
+        rmtI32 ivalue;
+        rmtF32 fvalue;
+    } value;
+    rmtU32 flags;
+    rmtU32 desc; // The hash of the description
```
JCash, line 4322: Code style Q: Do you prefer shorter names, or the names spelled out?
E.g.. `desc` vs `description` ?
`StatInfo`vs `StatisticInfo` ?

lib/Remotery.c
```diff
@@ -4309,6 +4311,17 @@ static void Server_Update(Server* server)
...
 #define SAMPLE_NAME_LEN 128
 
+typedef struct StatInfo
+{
+    enum rmtStatType type;
+    union {
+        rmtI32 ivalue;
+        rmtF32 fvalue;
+    } value;
+    rmtU32 flags;
```
JCash, line 4321: Controlling the behavior of the statistic.

lib/Remotery.c
```diff
@@ -8892,6 +8960,26 @@ RMT_API rmtSampleType _rmt_SampleGetType(rmtSample* sample)
...
+}
+
+// TODO: How to deal with getting wrong type?
+// SHould it return an rmtError instead?
+RMT_API rmtI32 _rmt_SampleGetStatValueI32(rmtSample* sample)
+{
+    if (sample->stat_info.type != RMT_StatType_I32)
+        return 0;
+    return sample->stat_info.value.ivalue;
+}
```
JCash, line 8980: Q: Should we use `rmtError` instead, or is it ok with returning 0.
I think keeping it simple with returning 0 should suffice.

dwilliamson:
This is a great start. I'll pull all my comments from the Issue in here as I think it's best to iterate on this.

Here's some quick points:

* Many counters will want to be modified from multiple threads.
   * This implies that stats being part of the sample hierarchy prevents that.
   * Any add/subtract/reset operations on a stat will need to be thread-safe (e.g. atomic, not locked).
   * Whenever I've profiled large systems with many cores, atomic update of stats has actually dented performance so I used a gather phase at the end of the frame to merge them.
   * Perhaps that's what can be done here: walk all threads looking for stats of the same name and sum them.
   * Summed stats can be placed in a "pseudo-thread" sampler that gets its own window for
   * This then allows us to optionally view stats as part of the thread sample hierarchy in one view and then we have the global view of summed stats which is its own window.
* Aggregate samples are slow.
   * They work by going up to the sample parent and then doing a linear search of all children for a sample with the same hash.
   * Counters/stats are going to be fairly frequent so I don't think we should rely on this tool.

To have global stats and work within the hierarchy I think we need two things:

* Define stats globally with a macro.
   * Each global stat creates a stat sample in the global Remotery instance - it will probably need its own Sample Tree.
   * Global stats can explicitly define parents as other global stats.
   * This gives us the opportunity to provide their default value, description and flags only once.
* When stats are modified they create samples on the current thread-sampler, just like you have now.
   * However, they are not aggregate and just modify the local stat.
   * They are aware of and point to their global parent stat since they share the same name.
      * Maybe they add themselves to a linked list in the parent stat to speed up the gather phase?
      * You can certainly do this in a thread-safe manner using existing Remotery CAS semantics.
* There is an end of frame gather phase on all the global stats.
   * Global stats walk through all their uses on all threads and sum.
   * Going to need some logic here managing stat resets.

The example that keeps going through my head with all this is a simple Triangles Drawn stat. With the above system we can define a global stat group named `Engine` and a global stat parented to that called `TrianglesDrawn`. The name ends up being `Engine.TrianglesDrawn`. 

Whenever an inner loop `DrawTriangles` function increments the counter, a new sample for `Engine.TrianglesDrawn` gets created and is parented to `Engine.TrianglesDrawn` in that call. This allows us optionally to inspect counters in the thread-view for how many triangles that single call has drawn.

Then the gather phase sums all of the `Engine.TrianglesDrawn` stats across all threads and presents that in a separate window using the same technique as your `delay.aggregate.MyCounter` view.

I am on Easter holidays from today but I'll still be able to review/approve; I just won't be able to modify any code.

lib/Remotery.c
dwilliamson: In the implementation file try to be as verbose as you can.

lib/Remotery.c
```diff
@@ -4376,6 +4391,8 @@ static rmtError Sample_Constructor(Sample* sample)
     sample->call_count = 0;
     sample->recurse_depth = 0;
     sample->max_recurse_depth = 0;
+    sample->stat_info.type = RMT_StatType_Count;
+    sample->stat_info.desc = 0;
```
dwilliamson, line 4395: Missing `flags` reset.

lib/Remotery.c
dwilliamson: 0 sounds good.

lib/Remotery.h
```diff
@@ -210,6 +210,8 @@ typedef unsigned char rmtU8;
 typedef unsigned short rmtU16;
 typedef unsigned int rmtU32;
 typedef unsigned long long rmtU64;
+typedef signed int rmtI32;
```
dwilliamson, line 213: We already have signed types below: `rmtS32`

JCash:
Yes I agree, if the same statistics are to be used over multiple threads, then the design needs to cater for this.

The idea of the global collection of statistics (e.g. a separate SampleTree), together with a gather step at the end of the frame, seems good.

I also like the linked list approach for gathering the statistics from all the threads when presenting the global statistics.
Doing the gather for the statistics of a single thread should be easy enough as they are then located within a single list of samples.

I think this setup will work well in presenting both global and local statistics.

I'll start working on this design, however I too am a bit occupied during the weekend, so I don't expect to get back on this until tuesday.

JCash:
Hi again @dwilliamson !

FInally got back to this task.
Here is an updated proof-of-concept to get closer to the ideas that you've previously outlined.
Hopefully I haven't strayed too far from your original ideas though. Feel free to pick it apart!

For the global statistics declarations, I chose to store them in a linked list (`root_stat`), since the ObjectAllocator has thread safety. For quick lookups, I also created a lookup table (name hash -> sample declaration) called `stat_decl_map`.

I chose to put the statistics in the regular CPU sample tree in order to preserve the hierarchy.
E.g. the comitted vertex count might come from various systems (sprites, gui, models, text etc)

When the sample tree is about to be consumed on the Remotery thread, I gather the stat data from it and apply it on the global stats. See example below for the end result.

A big question is when to send/reset the global statistics.
Currently, I set the "dirty" flag once another threads' statistics has been gathered into the global tree. But it feels a bit glitchy to be honest.

Global stats:

* When to present/reset the global sample tree?
* Use a fixed tick rate? (e.g. 60fps)

Per Thread:

* Currently a list of each stat invocation (i.e. not aggregated). See output below.
* How to visualize the final result?
    * Let the visualizer (javascript, client using iterator) handle it?



Current example:
`Remotery Stat Tree` is the global stat tree.
AggregateCallCount uses `rmt_StatAddS32()`, RecursiveCallDepth uses `rmt_StatSetS32()`

```
// ********************   DUMP TREE: Thread1   ************************
SAMPLE: delay 1  time: 46  self: 18 type: 0  color: 0xddf9f8
  SAMPLE: recursive 6  time: 18  self: 18 type: 0  color: 0x6e745f
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 0  desc: -
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 1  desc: -
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 2  desc: -
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 3  desc: -
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 4  desc: -
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 5  desc: -
  SAMPLE: aggregate 3  time: 10  self: 10 type: 0  color: 0xcfd6ff
    STAT: AggregateCallCount type: RMT_StatType_I32 value: 1  desc: -
    STAT: AggregateCallCount type: RMT_StatType_I32 value: 1  desc: -
    STAT: AggregateCallCount type: RMT_StatType_I32 value: 1  desc: -
// ********************   DUMP TREE: Remotery Stat Tree   ************************
SAMPLE: (null) 1  time: 0  self: 0 type: 5  color: 0x677777
  STAT: Sample type: RMT_StatType_Group desc: Top scope
    STAT: AggregateCallCount type: RMT_StatType_I32 value: 3  desc: Sums up all calls
    STAT: RecursiveCallDepth type: RMT_StatType_I32 value: 5  desc: Shows the max depth
```

dwilliamson:
Thanks, I'll look at this in detail this weekend.

JCash:
Hi again, just wanted to check if you have any initial thoughts about the second version?

JCash:
Regarding when to reset the global stats counters, I think an explicit sync point that the user can call is the clearest way.

E.g. at the start of the step function of the game's main thread, you might have something like this:

```
rmt_ResetCPUStats(); // Called once per game "frame"
rmt_BeginCPUSample(MainLoop);
rmt_EndCPUSample();
```

JCash:
Assuming the API won't change too much, I'm going to start to flesh out support for floats as well, and I'll add the `rmt_ResetCPUStats()`  (name is ofc pending review, renaming things isn't a problem).

I want to see this in action in our engine so I can have a WIP inmplementation in place and deliver a first version of it.

dwilliamson:
Thanks again for all your work and I'm sorry this is taking so long. I have so much work right now!

I started a review but realised this wasn't quite what I had in mind, so I just left the couple review comments and decided to focus more closely on this.

This is what I've come up with:

```
// we want properties that can be set and updated by the game
// and then properties that can be set by the UI itself
// mutex lock around the whole stats tree

#include <string.h>

typedef struct{
    struct rmtProperty* firstProperty;
    struct rmtProperty* lastProperty;
} Remotery2;

Remotery2 g_Remotery2;

typedef enum
{
    RMT_PropertyType_rmtGroup,
    RMT_PropertyType_rmtBool,
    RMT_PropertyType_rmtS32,
    RMT_PropertyType_rmtU32,
    RMT_PropertyType_rmtF32,
    RMT_PropertyType_rmtS64,
    RMT_PropertyType_rmtU64,
    RMT_PropertyType_rmtF64,
} rmtPropertyType;

typedef enum
{
    RMT_PropertyFlags_NoFlags = 0,
    RMT_PropertyFlags_FrameReset = 1,
} rmtPropertyFlags;

typedef struct
{
    int empty;
} rmtGroup;

typedef union rmtPropertyValue
{
    #ifdef __cplusplus
    rmtPropertyValue() { }
    // These are static Make calls, rather than overloaded constructors, because `rmtBool` is the same type as `rmtU32`
    static rmtPropertyValue MakeBool(rmtBool v) { rmtPropertyValue pv; pv.Bool = v; return pv; }
    static rmtPropertyValue MakeS32(rmtS32 v) { rmtPropertyValue pv; pv.S32 = v; return pv; }
    static rmtPropertyValue MakeU32(rmtU32 v) { rmtPropertyValue pv; pv.U32 = v; return pv; }
    static rmtPropertyValue MakeF32(rmtF32 v) { rmtPropertyValue pv; pv.F32 = v; return pv; }
    static rmtPropertyValue MakeS64(rmtS64 v) { rmtPropertyValue pv; pv.S64 = v; return pv; }
    static rmtPropertyValue MakeU64(rmtU64 v) { rmtPropertyValue pv; pv.U64 = v; return pv; }
    static rmtPropertyValue MakeF64(rmtF64 v) { rmtPropertyValue pv; pv.F64 = v; return pv; }
    #endif

    rmtBool Bool;
    rmtS32 S32;
    rmtU32 U32;
    rmtF32 F32;
    rmtS64 S64;
    rmtU64 U64;
    rmtF64 F64;
} rmtPropertyValue;

typedef struct rmtProperty
{
    rmtBool initialised;
    rmtPropertyType type;
    rmtPropertyValue value;
    rmtPropertyFlags flags;
    const char* name;
    const char* description;
    rmtPropertyValue defaultValue;
    struct rmtProperty* parent;

    struct rmtProperty* firstChild;
    struct rmtProperty* lastChild;
    struct rmtProperty* nextSibling;
} rmtProperty;

void _rmt_RegisterProperty(rmtProperty* property)
{
    if (property->initialised == RMT_FALSE)
    {
        rmtProperty* parent_property = property->parent;
        if (parent_property == NULL)
        {
            // his is a root property: add it to the remotery object to keep track of them all
            if (g_Remotery2.firstProperty == NULL)
            {
                g_Remotery2.firstProperty = property;
                g_Remotery2.lastProperty = property;
            }
            else
            {
                g_Remotery2.lastProperty->nextSibling = property;
                g_Remotery2.lastProperty = property;
            }
        }
        else
        {
            _rmt_RegisterProperty(parent_property);

            // Link this property into the parent's list
            if (parent_property->firstChild != NULL)
            {
                parent_property->lastChild->nextSibling = property;
                parent_property->lastChild = property;
            }
            else
            {
                parent_property->firstChild = property;
                parent_property->lastChild = property;
            }
        }

        // queue name for hash map adding
        
        // register property
        property->initialised = RMT_TRUE;
    }
}

void _rmt_PropertySetValue(rmtProperty* property)
{
    _rmt_RegisterProperty(property);

    // on this thread, create a new sample that encodes the value just set

    // send the sample to remotery UI and disk log

    // value resets and sets don't have delta values, really
}

void _rmt_PropertyAddValue(rmtProperty* property, rmtPropertyValue add_value)
{
    _rmt_RegisterProperty(property);

    // use `add_value` to determine how much this property was changed

    // on this thread, create a new sample that encodes the delta and parents itself to `property`
    // could also encode the current value of the property at this point

    // send the sample to remotery UI and disk log
}

#define _rmt_PropertyDefine(type, name, default_value, flags, desc, ...) \
    rmtProperty name = { RMT_FALSE, RMT_PropertyType_##type, default_value, flags, #name, desc, default_value, __VA_ARGS__ };

// C++ doesn't support designated initialisers until C++20
// Worth checking for C++ designated initialisers to remove the function call in debug builds
#ifdef __cplusplus
#define _rmt_MakePropertyValue(field, value) rmtPropertyValue::Make##field(value)
#else
#define _rmt_MakePropertyValue(field, value) { .field = value }
#endif

#define rmt_PropertyDefine_Group(name, desc, ...) _rmt_PropertyDefine(rmtGroup, name, _rmt_MakePropertyValue(Bool, 0), RMT_PropertyFlags_NoFlags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_Bool(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtBool, name, _rmt_MakePropertyValue(Bool, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_S32(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtS32, name, _rmt_MakePropertyValue(S32, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_U32(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtU32, name, _rmt_MakePropertyValue(U32, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_F32(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtF32, name, _rmt_MakePropertyValue(F32, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_S64(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtS64, name, _rmt_MakePropertyValue(S64, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_U64(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtU64, name, _rmt_MakePropertyValue(U64, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)
#define rmt_PropertyDefine_F64(name, default_value, flags, desc, ...) _rmt_PropertyDefine(rmtF64, name, _rmt_MakePropertyValue(F64, default_value), RMT_PropertyFlags_##flags, desc, __VA_ARGS__)

#define rmt_PropertyExtern(type, name) \
    extern rmtProperty name;

#define _rmt_PropertySet(field, name, set_value) \
    { \
        name.value.field = set_value; \
        _rmt_PropertySetValue(&name); \
    }

#define rmt_PropertySet_Bool(name, set_value) _rmt_PropertySet(Bool, name, set_value)
#define rmt_PropertySet_S32(name, set_value) _rmt_PropertySet(S32, name, set_value)
#define rmt_PropertySet_U32(name, set_value) _rmt_PropertySet(U32, name, set_value)
#define rmt_PropertySet_F32(name, set_value) _rmt_PropertySet(F32, name, set_value)
#define rmt_PropertySet_S64(name, set_value) _rmt_PropertySet(S64, name, set_value)
#define rmt_PropertySet_U64(name, set_value) _rmt_PropertySet(U64, name, set_value)
#define rmt_PropertySet_F64(name, set_value) _rmt_PropertySet(F64, name, set_value)

#define _rmt_PropertyAdd(field, name, add_value) \
    { \
        name.value.field += add_value; \
        rmtPropertyValue delta_value = _rmt_MakePropertyValue(field, add_value); \
        _rmt_PropertyAddValue(&name, delta_value); \
    }

#define rmt_PropertyAdd_S32(name, add_value) _rmt_PropertyAdd(S32, name, add_value)
#define rmt_PropertyAdd_U32(name, add_value) _rmt_PropertyAdd(U32, name, add_value)
#define rmt_PropertyAdd_F32(name, add_value) _rmt_PropertyAdd(F32, name, add_value)

#define rmt_PropertyReset(name) \
    { \
        name.value = name.defaultValue; \
        _rmt_PropertySetValue(&name); \
    }

void _rmt_PropertyFrameReset(rmtProperty* first_property)
{
    for (rmtProperty* property = first_property; property != NULL; property = property->nextSibling)
    {
        if (property->type == RMT_PropertyType_rmtGroup)
        {
            _rmt_PropertyFrameReset(property->firstChild);
        }

        else if ((property->flags & RMT_PropertyFlags_FrameReset) != 0)
        {
            property->value = property->defaultValue;
        }
    }
}

void _rmt_PropertyFrameResetAll()
{
    _rmt_PropertyFrameReset(g_Remotery2.firstProperty);
}

#define rmt_PropertyFrameResetAll() _rmt_PropertyFrameResetAll()

rmt_PropertyDefine_Group(Game, "Game Property Group");
rmt_PropertyDefine_Bool(Enabled, RMT_FALSE, NoFlags, "Enabled?", &Game);
rmt_PropertyDefine_S32(NbThings, 0, FrameReset, "Number of Things", &Game);
rmt_PropertyDefine_F32(SizeOfIt, 1.0f, NoFlags, "Look at the size of it!", &Game);

static void test()
{
    rmt_PropertySet_Bool(Enabled, RMT_TRUE);
    rmt_PropertySet_S32(NbThings, 1);
    rmt_PropertySet_F32(SizeOfIt, 1.1f);

    rmt_PropertyAdd_S32(NbThings, 3);
    rmt_PropertyAdd_F32(SizeOfIt, 4.5f);

    rmt_PropertyReset(NbThings);
    rmt_PropertyReset(SizeOfIt);

    rmt_PropertyFrameResetAll();
}
```

Some important points:

* I relealised one goal of Remotery is to provide you with a UI for runtime-configuration of your engine/game, achieved through engine properties you can expose. Stats a very similar and a subset of this functionality, hence the naming.
* Properties are global variables with easy raw access and modification.
* There will be future support for properties that are actual structures instead of native C types.
* You manually call `rmt_PropertyFrameResetAll` each frame.
* I didn't write the bit that creates samples and inserts into the tree but how that's achieved should be understandable from the comments.
* This compiles on both C and C++.
* Registration of properties and progressive building of the property tree will need to be mutex-locked. This isn't really a problem as it's only done once on first use of each property.
* It interacts seamlessly with the existing name hash system.

I'll focus some more on this tomorrow but your help in guiding this has been great.

JCash:
Thanks for the feedback!
I think for sure this design is preferable to forcing it to live inside the sample tree.

Since you explicitly mentioned registration of the property needing a lock, I wonder on your thoughts about locking the scope when setting/adding to the property?

Also, I think it's a good idea to do the send/reset of the properties in one call.
Perhaps a different name than `rmt_MarkFrameEnd()`? To me, The name implies the existence of a mark frame begin. (Incidentally, our own api actually has a begin/end frame function, which we used for this purpose as well).

> I'll focus some more on this tomorrow

Looking forward to your next iteration on the api.
Is this all pseudo code, or are you actually implementing parts of it?

dwilliamson:
I'm implementing part of this right now.

> Since you explicitly mentioned registration of the property needing a lock, I wonder on your thoughts about locking
> the scope when setting/adding to the property?

I want to leave this for a future revision but these could just be atomic operations. I'm not sure yet whether I want to make them atomic uncondtionally or whether the API needs modification (e.g. maybe we add `RMT_PropertyType_rmtU32Atomic`).

> Also, I think it's a good idea to do the send/reset of the properties in one call.
> Perhaps a different name than rmt_MarkFrameEnd()? To me, The name implies the existence of a mark
> frame begin. (Incidentally, our own api actually has a begin/end frame function, which we used for this purpose as well).

The end frame is worth some discussion as the RHIs I make these days only have an EndFrame. I began to realise that in order to make a thread-resilient API you need to accept function calls outside of any fixed range defined by a Begin/End Frame. As the implementations evolved, eventually all logic moved out of BeginFrame and into EndFrame.

But this just highlights the disparity and make me think you are right. Eventually my goals here are to expose my entire game's object hierarchy via properties. Obviously this won't work if we're sending all that data every frame so I'm imagining some sort of "object is dirty" API that only sends changes.

To avoid speculating too much into the future I think it may be wise to keep them as two functions for now: something like `rmt_PropertyFrameResetAll` and `rmt_PropertySnapshotAll`. They'll be next to each other in the public API and so hard to miss. though you'll need to call the snapshot before reset.

JCash:
That's great! I'll have a look & start using this in the engine to get a feel for it!

Regarding the API for traversing the properties, I see a few options.
Following the pattern of using callbacks for the sample trees:
* Set a callback in the settings file
* Pass in a callback to `rmt_PropertyFrameResetAll`

Although a getter function (`rmt_PropertyGetRoot()`) feels clean and intuitive,
it also has the issue of threading.

What do you think?

JCash:
I think setting the property callback in the Remotery settings is consistent with the current api.
Also, the call site of the `rmt_PropertyFrameResetAll()` might not be as suitable for that type of code (less code dependencies).

JCash:
I think this comment is a bit confusing:
"//    * `flags` are specified without the `RMT_PropertyFlags_` prefix."

Currently, the argument is called `flags`, but given the comment and the implementation, I suggest it's renamed to `flag`.

dwilliamson:
Ummm... yes, you couldn't specify multiple flags anyway because you'd then need to use the prefix, so it should be `flag`.

I'm open to iterating on this bit because I'm not even sure "frame reset" makes sense. You want to mark specific properties for reset when you call a function but you may not want to link it to "frames."

dwilliamson:
> I think setting the property callback in the Remotery settings is consistent with the current api.

Agreed.

JCash:
I've added a first version of the property iterators [here](https://github.com/Celtoys/Remotery/pull/197)

It has a design question regarding the property accessors.

JCash:
Since the implementation is well underway using the new api, I think it's time to close this PR.

PR #197, opened 2022-05-18
Property iterator

JCash:


lib/Remotery.h
```diff
@@ -530,21 +538,34 @@ typedef enum
...
+{
+    RMT_PropertyType_rmtGroup,
+    RMT_PropertyType_rmtBool,
+    RMT_PropertyType_rmtS32,
+    RMT_PropertyType_rmtU32,
+    RMT_PropertyType_rmtF32,
+    RMT_PropertyType_rmtS64,
+    RMT_PropertyType_rmtU64,
+    RMT_PropertyType_rmtF64,
+} rmtPropertyType;
```
JCash, line 562: Needed as public for iterators

lib/Remotery.h
```diff
@@ -587,19 +608,6 @@ typedef enum
...
-    RMT_PropertyType_rmtS32,
-    RMT_PropertyType_rmtU32,
-    RMT_PropertyType_rmtF32,
-    RMT_PropertyType_rmtS64,
-    RMT_PropertyType_rmtU64,
-    RMT_PropertyType_rmtF64,
-} rmtPropertyType;
-
 // A property value as a union of all its possible types
 typedef union rmtPropertyValue
```
JCash, line 622: Q: SHould we make this public as well, or add multiple getter functions to the property api:
`rmt_PropertyGetS32()` vs `rmt_PropertyGetValue()`

sample/dump.c
```diff
@@ -72,12 +83,56 @@ void dumpTree(void* ctx, rmtSampleTree* sample_tree)
...
     printTree(root, 0);
 }
 
+void printProperty(rmtProperty* property, int indent)
+{
+    rmtPropertyIterator iter;
+
+    const char* name = rmt_PropertyGetName(property);
+    rmtPropertyType type = rmt_PropertyGetType(property);
+    rmtPropertyValue value = rmt_PropertyGetValue(property);
```
JCash, line 102: The iterator use case:
Should we get the union value, or use specific getters for each type?

lib/Remotery.h
```diff
@@ -220,6 +220,19 @@ typedef struct Sample rmtSample;
...
+
+typedef enum rmtSampleType
+{
+    RMT_SampleType_CPU,
+    RMT_SampleType_CUDA,
+    RMT_SampleType_D3D11,
+    RMT_SampleType_OpenGL,
+    RMT_SampleType_Metal,
+    RMT_SampleType_Count,
+} rmtSampleType;
```
JCash, line 234: Missed this when we added support for sample iterators.
This enum is used when iterating the samples.

lib/Remotery.h
dwilliamson: I am happy to make `rmtProperty` and `rmtPropertyValue` public and maintain their API.

It keeps everything super simple.

I've been currently pondering on the idea of making `Sample` public, as well. Not sure about that one yet...

sample/dump.c
dwilliamson: Make `rmtPropertyValue` public and let the user access that based on type.

lib/Remotery.c
```diff
@@ -9284,8 +9345,13 @@ RMT_API void _rmt_PropertyFrameResetAll()
     {
         return;
     }
-        
+
     mtxLock(&g_Remotery->propertyMutex);
+    if (g_Settings.property_handler != NULL)
```
dwilliamson, line 9350: I think `_rmtPropertySnapshotAll` is a better place for this as that's when the data gets sent to the viewer and logged. Potentially that means that the callback should be referenced as a "snapshot callback" as well.

JCash: I agree. It was a case of me reading poorly and not seeing that the function existed 🙄

lib/Remotery.h
```diff
@@ -654,6 +617,62 @@ typedef struct rmtProperty
     rmtU32 nameHash;
 } rmtProperty;
 
+// Define properties of different types at global scope:
```
JCash, line 620: I didn't move this, but the rmtProperty/rmtPropertyValue.

sample/dump.c
```diff
@@ -112,6 +170,12 @@ int main() {
         rmt_LogText("start profiling");
         delay();
         rmt_LogText("end profiling");
+
+        rmt_PropertySet_Bool(WasUpdated, RMT_TRUE);
+        rmt_PropertyAdd_U32(FrameCounter, 1);
+
+        rmt_PropertySnapshotAll();
```
JCash, line 177: @dwilliamson : ⚠️ I haven't figured out the actual issue yet, but when adding this call, the remotery destructor fails  on `Delete(ObjectAllocator, rmt->propertyAllocator);`:
 ```
 Assertion failed: (allocator->nb_inuse == 0), function ObjectAllocator_Destructor, file lib/Remotery.c, line 2132.
 ```
 
The count is 6, which I assume might be the root property + my 5 properties in dump.c.

Tested with:
```
$ clang -std=c89 -DRMT_ENABLED=1 -DRMT_ARCH_64BIT -g -O0 -Ilib sample/dump.c lib/Remotery.c -o test && ./test
```

JCash: I believe it's because the snapshot data is still in flight, and we're shutting down the Remotery instance before that call is handled.

dwilliamson: Hmmm... `BIN_ERROR_CHECK` will leak memory in `Remotery_SendPropertySnapshot` if if fails any write, not calling `FreePropertySnapshots`. Will have to pull the serialisation into a separate function.

dwilliamson: Found the shutdown leak, as well...

PR #204, opened 2022-05-20
Fix for rmtTryNew when __va_args_ is empty

JCash:


lib/Remotery.c
```diff
@@ -769,12 +769,25 @@ static rmtU32 GaloisLFSRNext(rmtU32 value, rmtU32 xor_mask)
...
     }
 
-// New will allocate enough space for the object and call the constructor
-// If allocation fails the constructor won't be called
-// If the constructor fails, the destructor is called and memory is released
-// NOTE: Use of sizeof() requires that the type be defined at the point of call
-// This is a disadvantage over requiring only a custom Create function
-#define rmtTryNew(type, obj, ...)                              \
+
+#define _VA_ARGS_SELECT(PREFIX,_5,_4,_3,_2,_1,SUFFIX,...) PREFIX ## _ ## SUFFIX
```
JCash, line 773: Inspired by info from this thread
https://stackoverflow.com/questions/5588855/standard-alternative-to-gccs-va-args-trick

lib/Remotery.c
```diff
@@ -789,6 +802,14 @@ static rmtU32 GaloisLFSRNext(rmtU32 value, rmtU32 xor_mask)
         }                                                      \
     }
 
+// New will allocate enough space for the object and call the constructor
+// If allocation fails the constructor won't be called
+// If the constructor fails, the destructor is called and memory is released
+// NOTE: Use of sizeof() requires that the type be defined at the point of call
+// This is a disadvantage over requiring only a custom Create function
+#define rmtTryNew(...)                                         \
+    _VA_ARGS_SELECT(_rmtTryNew,__VA_ARGS__,N,N,N,1,1)(__VA_ARGS__)
```
JCash, line 811: It almost compiles now.
The remaining issue is that not all functions have or return an `error` value.

PR #216, opened 2022-06-17
Custom hash function

JCash:
In our use case, I'd like to store the same hashes for later presenting in our internal UI.
This allows me to avoid going back and forth between strings and our own hash functions.

By using our internal hash function, I can keep the minimal storage needed for each recorded frame.

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -3537,7 +3537,7 @@ static rmtU32 MurmurHash3_x86_32(const void* key, int len, rmtU32 seed)
     return h1;
 }
 
-rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
+RMT_API rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
 {
     return MurmurHash3_x86_32(s, len, seed);
 }
@@ -3546,7 +3546,7 @@ rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
     #if defined(__cplusplus)
     extern "C"
     #endif
-    rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed);
+    RMT_API rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed);
 
 #endif // RMT_USE_INTERNAL_HASH_FUNCTION
 
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -9863,7 +9863,7 @@ static void RegisterProperty(rmtProperty* property, rmtBool can_lock)
 
             // Calculate the name hash and send it to the viewer
             name_len = strnlen_s(property->name, 256);
-            property->nameHash = MurmurHash3_x86_32(property->name, name_len, 0);
+            property->nameHash = rmt_HashString32(property->name, name_len, 0);
             QueueAddToStringTable(g_Remotery->mq_to_rmt_thread, property->nameHash, property->name, name_len, NULL);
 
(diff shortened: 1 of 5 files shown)


lib/Remotery.c
```diff
@@ -3450,6 +3450,8 @@ static void Base64_Encode(const rmtU8* in_bytes, rmtU32 length, rmtU8* out_bytes
 // domain. The author hereby disclaims copyright to this source code.
 //-----------------------------------------------------------------------------
 
+#if RMT_USE_INTERNAL_HASH_FUNCTION
```
JCash, line 3453: I've tried to copy the naming and usage style of the other options in the header file.

Not sure about the name. Is it clear enough?

lib/Remotery.c
```diff
@@ -3535,6 +3537,19 @@ static rmtU32 MurmurHash3_x86_32(const void* key, int len, rmtU32 seed)
...
+rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
+{
+    return MurmurHash3_x86_32(s, len, seed);
+}
+
+#else
+    #if defined(__cplusplus)
+    extern "C"
+    #endif
+    rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed);
```
JCash, line 3549: I opted for avoiding setting a custom function pointer.
Given the other compile time settings, it made the most sense.
Also, I've in favor of removing unused code.

dwilliamson:
Clever idea. I like this instead of function pointers.

lib/Remotery.h
```diff
@@ -109,6 +109,11 @@ documented just below this comment.
 #define RMT_D3D11_RESYNC_ON_DISJOINT 1
 #endif
 
+// If RMT_USE_INTERNAL_HASH_FUNCTION is defined, the internal function is implemented.
```
dwilliamson, line 112: Perhaps...

```
// If RMT_USE_INTERNAL_HASH_FUNCTION is defined to 1, the internal hash function for strings is used.
// This is the default setting.
// If you set RMT_USE_INTERNAL_HASH_FUNCTIOn to 0, you must implement rmt_HashString32 yourself.
```

lib/Remotery.h
```diff
@@ -729,6 +734,15 @@ typedef struct rmtProperty
 // Reset all RMT_PropertyFlags_FrameReset properties to their default value
 #define rmt_PropertyFrameResetAll() _rmt_PropertyFrameResetAll()
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed);
```
dwilliamson, line 741: Move this to the private section right at the bottom.

I'll move all the private property stuff down later as well, when I get over my current task.

JCash: As a private function, should it also be prefixed with "_"?

dwilliamson: Hmmm... yes.

lib/Remotery.c
```diff
@@ -3537,7 +3537,7 @@ static rmtU32 MurmurHash3_x86_32(const void* key, int len, rmtU32 seed)
     return h1;
 }
 
-rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
+RMT_API rmtU32 rmt_HashString32(const char* s, int len, rmtU32 seed)
```
dwilliamson, line 3540: Should this be `_rmt_HashString32`?

JCash: It should!

PR #222, opened 2022-07-08
Missing include for FD_ZERO

JCash:


lib/Remotery.c
```diff
@@ -147,6 +147,7 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
         #include <pthread.h>
         #include <unistd.h>
         #include <string.h>
+        #include <sys/select.h>
```
JCash, line 150: Noticed this on one of the console platforms

JCash:
Not entirely sure why the CI builds fail 🤔

dwilliamson:
Spurious errors from `echo off` call.

Issue #223, opened 2022-07-17
Apps targeting Android 10 cannot directly use ashmem (/dev/ashmem)


AGulev:
Android 10 and higher can't use `/dev/ashmem` directly anymore and must use NDK's `ASharedMemory` class.
Because of that Remotery doesn't work on Android 10+.

More info is available here:
https://stackoverflow.com/a/53062250

JCash:
The new Android api does seem to still use the `ashmem_create_region() [internally](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/native/android/sharedmem.cpp#53), but it is a lot more involved, and it currently uses `std::string` for I/O:
https://android.googlesource.com/platform/system/core/+/refs/heads/master/libcutils/ashmem-dev.cpp#365

Perhaps dynamically loading the shared library and grabbing the function pointer might be an option 🤔

AGulev:
Yes,
This is how google chromium does that
https://chromium.googlesource.com/chromium/src/+/refs/heads/main/base/android/linker/modern_linker_jni.cc#186

PR #224, opened 2022-07-18
Added support for Android 10+ devices

JCash:
Fixes https://github.com/Celtoys/Remotery/issues/223

Also added Android build for easier testing on CI.

### DIFF
--- a/.github/scripts/android_env.sh
+++ b/.github/scripts/android_env.sh
@@ -12,28 +12,6 @@ echo "ANDROID_HOME=${ANDROID_HOME}"
 echo "ANDROID_NDK_HOME=${ANDROID_NDK_HOME}"
 echo "ANDROID_NDK_API_VERSION=${ANDROID_NDK_API_VERSION}"
 
-# if [ ! -d "${ANDROID_NDK_HOME}" ]; then
-#     echo "No such directory ANDROID_NDK_HOME='${ANDROID_NDK_HOME}'"
-#     exit 1
-# fi
-
-echo "ANDROID_HOME:"
-ls -la ${ANDROID_HOME}
-
-echo "--------------------------"
-
-echo "ANDROID_HOME/ndk:"
-ls -la ${ANDROID_HOME}/ndk
-
-echo "--------------------------"
-
-echo "ANDROID_NDK_HOME:"
-
-ls -la ${ANDROID_NDK_HOME}
-
-echo "--------------------------"
-
-
 toolchainhost="linux"
 if [ "$(uname)" == "Darwin" ]; then
     toolchainhost="darwin"
--- a/.github/scripts/android_env.sh
+++ b/.github/scripts/android_env.sh
@@ -2,11 +2,6 @@
 
 set -e
 
-# Grab the ndk major version
-if [ "" != "${ANDROID_NDK_VERSION}" ]; then
-    ANDROID_NDK_API_VERSION=$(echo "${ANDROID_NDK_VERSION}"| cut -f1 -d'.')
-fi
-
 if [ "" == "${ANDROID_NDK_API_VERSION}" ]; then
     ANDROID_NDK_API_VERSION='16' # Android 4.1
 fi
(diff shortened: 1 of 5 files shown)


.github/scripts/android_env.sh
```diff
@@ -0,0 +1,43 @@
...
+
+echo "ARCH=${ARCH}"
+
+toolchainarch="armv7a-linux-androideabi${ANDROID_NDK_API_VERSION}"
+if [ "${ARCH}" == "arm64" ]; then
+    toolchainarch="aarch64-linux-android${ANDROID_NDK_API_VERSION}"
+fi
+
+export GCC=${ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${toolchainhost}-x86_64/bin/${toolchainarch}-clang
+export GXX=${ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${toolchainhost}-x86_64/bin/${toolchainarch}-clang++
```
JCash, line 28: There are no GCC for android anymore, so we're using clang here as well.

.github/scripts/compile.bat
```diff
@@ -1,7 +1,8 @@
 echo off
 
 set PLATFORM=%1
-set BUILDMODE=%2
+set ARCH=%2
+set BUILDMODE=%3
```
JCash, line 5: Added platform argument

.github/scripts/compile.sh
```diff
@@ -15,48 +19,79 @@ function run {
...
 
-
 LINKFLAGS="-lm -pthread"
-if [ "$(uname)" == "Darwin" ]; then
+if [ "macos" == "${PLATFORM}" ]; then
     LINKFLAGS=""
 fi
 
+if [ "android" == "${PLATFORM}" ]; then
+    . ${SCRIPT_DIR}/android_env.sh
```
JCash, line 39: Enable setting platform specific settings.

.github/workflows/build.yml
```diff
@@ -48,6 +56,40 @@ jobs:
...
+
+  build-android:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        platform: ['android']
+        arch: ['armv7', 'arm64']
+        buildmode: ['release']
+        ndkversion: [21]
```
JCash, line 71: So multiple versions could be tested

.github/workflows/build.yml
```diff
@@ -48,6 +56,40 @@ jobs:
...
+
+      - name: Set up JDK 1.11
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu' # See 'Supported distributions' for available options
+          java-version: '11'
+
+      # Sets ANDROID_HOME as well as installs the latest NDK
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v2
```
JCash, line 84: It always installs the latest sdk+ndk

.github/workflows/build.yml
```diff
@@ -48,6 +56,40 @@ jobs:
...
+
+      - name: Build
+        run: .github/scripts/compile.sh ${{ matrix.platform }} ${{ matrix.arch }} ${{ matrix.buildmode }}
+        env:
+          ANDROID_NDK_HOME: ${{ env.ANDROID_HOME }}/ndk-bundle
+          ANDROID_NDK_API_VERSION: ${{ matrix.ndkversion }}
+
+      # - name: Run
+      #   run: .github/scripts/run_tests.sh
```
JCash, line 93: Eventually, we want to be able to run the tests in an emulator as well.

lib/Remotery.c
```diff
@@ -1012,6 +1012,45 @@ int ashmem_create_region(const char* name, size_t size)
     close(fd);
     return ret;
 }
+
+// https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/ashmem/ashmem-dev.c
```
JCash, line 1016: As seen in Chromium.
Tested on Android 8 and Android 11

dwilliamson: Can we have a copy of the license here as well? AFAIA the Apache License 2.0 requires it.

dwilliamson: Also, use the following link so that it persists if `HEAD` is changed and it links directly to the line: https://chromium.googlesource.com/chromium/src/+/ad02487d87120bd66045960ffafe0fc27600af50/third_party/ashmem/ashmem-dev.c#181

dwilliamson (commented):
This is great stuff; good to see.

PR #225, opened 2022-07-19
Reset all properties in the tree

JCash:


lib/Remotery.c
```diff
@@ -10086,12 +10086,13 @@ static void PropertyFrameReset(Remotery* rmt, rmtProperty* first_property)
     rmtProperty* property;
     for (property = first_property; property != NULL; property = property->nextSibling)
     {
+        PropertyFrameReset(rmt, property->firstChild);
```
JCash, line 10089: Since it's currently possible to add properties to other properties (not just groups), then I think this change makes sense.

But, I don't think we've explicitly said it's in alignment with the design?

Personallty, I think it's nice to use that real estate in the visualisations to actually have a value there.
E.g my use case:

```
Cmponents
    Sprites           7
        VertexCount   28
        VertexSize    560
    ParticleFx        0
        VertexCount   0
        VertexSize    0
    ...

```

dwilliamson: Oh, crikey, no... I was not expecting this use-case.

That's a super clever idea.

PR #232, opened 2022-08-12
Moved potential side effect from assert()

JCash:


lib/Remotery.c
```diff
@@ -4226,10 +4226,13 @@ static Message* rmtMessageQueue_AllocMessage(rmtMessageQueue* queue, rmtU32 payl
...
 static void rmtMessageQueue_CommitMessage(Message* message, MessageID id)
 {
+    MessageID r;
     assert(message != NULL);
 
     // Setting the message ID signals to the consumer that the message is ready
-    assert(LoadAcquire((rmtU32*)&message->id) == MsgID_NotReady);
+    r = (MessageID)LoadAcquire((rmtU32*)&message->id);
+    (void)r;
+    assert(r == MsgID_NotReady);
```
JCash, line 4235: In the case someone builds with NDEBUG, removing all the asserts.

dwilliamson:
Any updates on this?

JCash:
Hmm, not sure what you mean?

I think it's a good rule of thumb to not include code with side effects in the assert macros, as they are removed when compiling with NDEBUG.
Although it may not be actually crucial in this case, it sets an example to others to not do it either.

dwilliamson:
The use of `(void)r` to notify a potentially unused parameter is not compatible across compilers. That's where we can use `RMT_UNREFERENCED_PARAMETER(r)`.

PR #238, opened 2022-09-07
fix TSAN warnings on macOS

conversy:


lib/Remotery.c
```diff
@@ -621,31 +622,42 @@ static void mtxDelete(rmtMutex* mutex)
...
+typedef _Atomic(rmtU64) rmtAtomicU64;
+typedef _Atomic(void*) rmtAtomicPtr;
+#else
 typedef volatile rmtS32 rmtAtomicS32;
 typedef volatile rmtU32 rmtAtomicU32;
 typedef volatile rmtU64 rmtAtomicU64;
+#endif
 
-static rmtBool AtomicCompareAndSwapU32(rmtU32 volatile* val, long old_val, long new_val)
+static rmtBool AtomicCompareAndSwapU32(rmtAtomicU32 volatile* val, rmtU32 old_val, rmtU32 new_val)
```
dwilliamson, line 636: Is there any reason you're preserving the `volatile` in this change? See `AtomicCompareAndSwapU64`.

conversy: no.
I thought it might be useful on other platforms to keep it this way, but apart from that, `volatile` might be removed altogether...

dwilliamson: No, if it compiles fine on Mac it'll be good. I'll change all that on my end.

dwilliamson:
Will merge this tomorrow when I get back in the office. Thanks...

lib/Remotery.c
```diff
@@ -2347,13 +2365,13 @@ static ObjectLink* ObjectAllocator_Pop(ObjectAllocator* allocator)
...
         {
             return NULL;
         }
-        ObjectLink* next_link = old_link->next;
-        if (AtomicCompareAndSwapPointer((long* volatile*)&allocator->first_free, (long*)old_link, (long*)next_link) ==
+        rmtAtomicPtr next_link = ((ObjectLink*) (old_link))->next;
+        if (AtomicCompareAndSwapPointer(&old_link, old_link, next_link) ==
             RMT_TRUE)
         {
             link = old_link;
```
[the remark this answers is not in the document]
conversy: I tried to follow the same logic as the previous code.

In the previous example, line 2333 performs an unprotected read which would trigger a warning:
```ObjectLink* old_link = (ObjectLink*)allocator->first_free;```

Then line 2335 casts the first argument into a `volatile` one:
```if (AtomicCompareAndSwapPointer((long* volatile*)&allocator->first_free, (long*)old_link, (long*)start) ==```

that will be used in:
```static rmtBool AtomicCompareAndSwapPointer(long* volatile* ptr, long* old_ptr, long* new_ptr)```

For the code I proposed, I changed those types to proper Atomic types.

In my understanding, this would require to change the type of the field of the structures (e.g. ``struct ObjectLink_s* volatile next;``) to a `rmtAtomicPtr*`. However, I decided to try the previous approach of first casting it to a pseudo atomic (i.e. `volatile`, instead this time it's a proper atomic type). This removed the warnings, but I'm not sure of the semantics.

A better approach might be to change the type of the fields for which we want to protect read and write with multiple threads?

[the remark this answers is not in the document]
JCash: But, at the point of AtomicCompareAndSwapPointer, the `allocator->first_free` may have changed, and you're passing in the address of a pointer allocated on the stack? It looks strange to me.

JCash:
I've updated my [draft PR](https://github.com/Celtoys/Remotery/pull/239) with a similar change as the one suggested here.
I've tested with asan, ubsan and tsan and it seems to behave correctly in the browser view.
Currently only tested on macOS.
My goal is to test it on Android as well (that's where I have some issues currently)

dwilliamson:
Thanks for submitting this. Going to move to the more complete https://github.com/Celtoys/Remotery/pull/239 based on some of your ideas.

conversy:
Sure, looking forward to it!

PR #239, opened 2022-09-09
Added initial version of thread.c

JCash:


JCash:
Compiles with 
```
clang -std=c89 -DRMT_ENABLED=1 -g -O0 -fsanitize=thread -Ilib sample/thread.c lib/Remotery.c -o build/test && ./build/test
```

dwilliamson:
So just to be clear: this defeats all your TSAN warnings, right? I will comb through this tomorrow but given we don't have load/acquire and store/release memory fences on non-x64/86 platforms, I feel this is only just the beginning...

lib/Remotery.c
```diff
@@ -2263,6 +2293,12 @@ typedef struct ObjectLink_s
     struct ObjectLink_s* volatile next;
 } ObjectLink;
 
+#if defined(RMT_USE_C_ATOMICS)
+    typedef _Atomic(ObjectLink*)    rmtAtomicObjectLinkPtr;
```
dwilliamson, line 2297: Nice. I like this. Maybe we can go one better and define a `rmtAtomicPtr(ObjectLink)` macro?

lib/Remotery.c
```diff
@@ -2137,7 +2167,7 @@ struct Thread_t
     rmtError error;
 
     // External threads can set this to request an exit
-    volatile rmtBool request_exit;
+    rmtAtomicU32 request_exit;
```
dwilliamson, line 2170: `rmtAtomicBool` that's a `_Atomic(rmtBool)` ?

lib/Remotery.c
```diff
@@ -4266,7 +4302,7 @@ Message* rmtMessageQueue_PeekNextMessage(rmtMessageQueue* queue)
     // the next one in the queue is ready.
     r = r & (queue->size - 1);
     ptr = (Message*)(queue->data->ptr + r);
-    id = (MessageID)LoadAcquire((rmtU32*)&ptr->id);
+    id = (MessageID)LoadAcquire((rmtAtomicU32*)&ptr->id);
```
dwilliamson, line 4305: Does this compile on C++? I can't see a `LoadAcquire` that matches the pointer type.

JCash: Well, it compiled as before but ran with asan warnings.
I now also added the stdatomic.h for c++11.

JCash: So now, with
```
clang -std=c11
clang++ -std=c++11
```
, there aren't any TSAN warnings.

With `clang -std=c99`, there are.
For C++ I don't get any warnings with `clang++ -std=c++98` either.

lib/Remotery.c
```diff
@@ -167,6 +167,10 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
     #include <cuda.h>
 #endif
 
+#if defined(RMT_USE_C_ATOMICS)
+    #include <stdatomic.h>
+#endif
+
```
JCash, line 173: I removed the compiler checks for this define, as it seems like both linux (gcc) and windows (cl) doesn't have it implemented?

JCash: Looking further into the build error on [linux/gcc](https://github.com/JCash/Remotery/actions/runs/3043210148/jobs/4902226495):
`error: ‘_Atomic’ does not name a type`

One suggested way to solve it for C++ would be:
```
#ifndef __cplusplus
# include <stdatomic.h>
#else
# include <atomic>
# define _Atomic(X) std::atomic< X >
#endif
```

Not sure how I feel about it, but I think we'd need something like this for compilers that implement stdatomic.h this way.

Benefit is that then the `RMT_USE_C_ATOMICS` can be set using the C11/C++11 check again.

lib/Remotery.c
```diff
@@ -6159,6 +6206,14 @@ static Remotery* g_Remotery = NULL;
 //
 static rmtBool g_RemoteryCreated = RMT_FALSE;
 
+static void rmtGetThreadNameFallback(char* out_thread_name, rmtU32 thread_name_size)
+{
+    // In cases where we can't get a thread name from the OS
+    out_thread_name[0] = 0;
+    strncat_s(out_thread_name, thread_name_size, "Thread", 6);
+    itoahex_s(out_thread_name + 6, thread_name_size - 6, AtomicAddU32(&g_Remotery->countThreads, 1));
```
JCash, line 6214: Moved since it now uses g_Remotery

lib/Remotery.c
```diff
@@ -2313,10 +2347,10 @@ static void ObjectAllocator_Destructor(ObjectAllocator* allocator)
     // Destroy all objects released to the allocator
     while (allocator->first_free != NULL)
     {
-        ObjectLink* next = allocator->first_free->next;
+        ObjectLink* next = ((ObjectLink*)allocator->first_free)->next;
```
JCash, line 2350: the `std::atomic` doesn't have a `->` operator, but a [cast operator](https://en.cppreference.com/w/cpp/atomic/atomic/operator_T).

lib/Remotery.c
dwilliamson: Oh I see, it's because there are no atomics on the C++ version.

lib/Remotery.c
```diff
@@ -6146,6 +6182,8 @@ struct Remotery
 
     // Frame used to determine age of property changes
     rmtU32 propertyFrame;
+
+    rmtAtomicS32 countThreads;
```
dwilliamson, line 6201: Nice.

lib/Remotery.c
```diff
@@ -167,6 +167,18 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
     #include <cuda.h>
 #endif
 
+#if __STDC_VERSION__ >= 201112L || __cplusplus >= 199711L
+    #if !defined(__STDC_NO_ATOMICS__)
+        #if !defined(RMT_USE_C_ATOMICS) // Check if the user already specified it
+            #define RMT_USE_C_ATOMICS
+        #endif
+    #endif
+#endif
```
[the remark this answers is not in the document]
dwilliamson: I guess we do because we want to have an explicit test case that ensures that path always compiles.

lib/Remotery.c
dwilliamson: This is totally not how I would do it, as I would have added separate `RMT_USE_C11_ATOMICS` and `RMT_USE_CPP_ATOMICS`.

But I have to admit: this way saves adding a whole bunch of C++ cases. And it actually seems to work.

For now, I think this is good enough and if we need to make it more succinct in future, we can easily do so.

lib/Remotery.c
dwilliamson: Ideally we'd have a weak load here but we don't have a primitive for that.

lib/Remotery.c
JCash: Yeah, and it didn't really work cleanly either.
So the latest version is playing with these two different defines.

dwilliamson:
Your line numbers are different to the version I have so I'm not sure what it's referencing.

At a guess I'd say it's the `hash_cache` read/write. Which makes sense: multiple threads will be reading and writing the cached hash value. However it's completely harmless as the value is either zero or the same calculated hash value across all threads.

The side effects are:

1) The string gets added to the string table more than once (which it can safely ignore).
2) The expensive path of calculating the hash happens more than once, which again is only once per thread in contention.

Options:

* Expose atomic types outside the implementation and use a `rmtAtomicU32` pointer, but I really don't like that.
* Can we cast to `rmtAtomicU32`?
* Put the whole thing in a CAS loop to ensure 100% validity.
* Can we comment the code in a way that TSAN classifies it as a false positive?

lib/Remotery.c
```diff
@@ -751,6 +751,32 @@ static void AtomicSubS32(rmtAtomicS32* value, rmtS32 sub)
...
+static rmtU32 AtomicLoadU32(rmtAtomicU32* value)
+{
+#if defined(RMT_USE_C11_ATOMICS)
+    return atomic_load(value);
+#elif defined(RMT_USE_CPP_ATOMICS)
+    return value->load();
+#elif defined(RMT_PLATFORM_WINDOWS) && !defined(__MINGW32__)
+    return (rmtU32)_InterlockedExchangeAdd((long volatile*)value, (long)0);
+#elif defined(RMT_PLATFORM_POSIX) || defined(__MINGW32__)
+    return (rmtU32)__sync_fetch_and_add(value, 0);
```
JCash, line 776: Couldn't find a "load" for these two, so I added `0` instead.

lib/Remotery.c
```diff
@@ -167,6 +167,28 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
     #include <cuda.h>
 #endif
 
+#if RMT_USE_LEGACY_ATOMICS==0
+    #if !defined(RMT_PLATFORM_WINDOWS)
```
JCash, line 171: Hopefully can enable this once the windows tests start working again.

lib/Remotery.c
```diff
@@ -621,33 +643,64 @@ static void mtxDelete(rmtMutex* mutex)
...
+#if defined(RMT_USE_C11_ATOMICS)
+    typedef _Atomic(rmtS32)     rmtAtomicS32;
+    typedef _Atomic(rmtU32)     rmtAtomicU32;
+    typedef _Atomic(rmtU64)     rmtAtomicU64;
+    typedef _Atomic(rmtBool)    rmtAtomicBool;
+    #define rmtAtomicPtr(type)  _Atomic(type *)
+#elif defined(RMT_USE_CPP_ATOMICS)
+    typedef std::atomic< rmtS32 >   rmtAtomicS32;
+    typedef std::atomic< rmtU32 >   rmtAtomicU32;
+    typedef std::atomic< rmtU64 >   rmtAtomicU64;
```
JCash, line 655: Strangely got a compiler error if I used the builtin atomic_uint64_t. It seemed to somehow mix in an "unsigned long" as opposed to an "unsigned long long" somewhere. In any case, this is good too, since we make sure it's the same base types.

lib/Remotery.h
```diff
@@ -116,6 +116,12 @@ documented just below this comment.
 #define RMT_USE_INTERNAL_HASH_FUNCTION 1
 #endif
 
+// If RMT_USE_LEGACY_ATOMICS is defined to 1, the implementation will use the legacy fallback atomic functions
+// The default setting is 0
+#ifndef RMT_USE_LEGACY_ATOMICS
+#define RMT_USE_LEGACY_ATOMICS 0
+#endif
```
JCash, line 123: New define:
* is the name ok?
* is the usage ok? i.e. setting/comparing a value, which is a bit different from other defines used.

.github/scripts/compile.bat
```diff
@@ -55,6 +55,7 @@ cl.exe %CL_OPTS% -Ilib -DRMT_USE_INTERNAL_HASH_FUNCTION=1 sample/sample.c lib/Re
 echo "Samples"
 cl.exe %CL_OPTS% -Ilib sample/sample.c lib/Remotery.c /link /out:build\sample.exe || goto :error
 cl.exe %CL_OPTS% -Ilib sample/dump.c lib/Remotery.c /link /out:build\dump.exe || goto :error
+cl.exe %CL_OPTS% -TP /MTd /O1 /fsanitize=address /Zi -Ilib sample/thread.c lib/Remotery.c /link /DEBUG /out:build\thread_asan.exe || goto :error
```
JCash, line 58: Trying out address sanitizer for windows

lib/Remotery.c
dwilliamson: We use `LoadAcquire` and `StoreRelease` for these operations.

Unfortunately TSAN does not work with these functions as it can't recognise manual memory barriers. But then: we have no manual memory barriers in there anyway, just compiler ones.

I am happy with this change for now, but we need to look forward to merging them and specifying memory order.

lib/Remotery.h
dwilliamson: This is exactly how we do stuff like `RMT_USE_D3D12` so good.

dwilliamson:
Is this ready to go? What is the status of running ASAN/UBSAN/TSAN after this change?

JCash:
I think it's good to go.
I have no asan/tsan/ubsan issues locally, and the linux/darwin tests produce no warnings either. And the windows asan test is green as well.

dwilliamson:
Awesome work, thanks!

PR #241, opened 2022-09-10
Add compiler flags from gcc and clang to determine bit depth of executable

JCash:


lib/Remotery.h
```diff
@@ -141,6 +141,14 @@ documented just below this comment.
 #endif
 #endif
 
+#if __GNUC__ || __clang__
+#if __x86_64__ || __ppc64__ || __amd64__
+#define RMT_ARCH_64BIT
+#else
+#define RMT_ARCH_32BIT
+#endif
+#endif
```
JCash, line 150: It is easy to forget to set this when compiling the samples.
Perhaps we should also check that either of them is actually set before continuing?

JCash: E.g.
```
#if !(defined(RMT_ARCH_64BIT) || defined(RMT_ARCH_32BIT)
    #error "No RMT_ARCH_64BIT or RMT_ARCH_32BIT is set!"
#endif
```

dwilliamson: I don't want to require these flags exist from the user because it will break years of legacy dependencies. The idea was that those flags should be identified from the compiler as much as is possible. If they're not set then safe fallbacks should be used.

JCash: Ok, I'll remove the check then.

Would it be better to output a warning?

I've forgot to set it more than once, and then the UBSAN will report a lot of issues since the code mixes and matches 32- and 64-bit code.

dwilliamson: Right now this PR looks good to go and the only `Remotery.c` dependency is the thread profiler, which is Windows-guarded now, anyway.

Would you still like to add a warning?

JCash: Nah, it's ok. I think I will perhaps add some note in the readme later on :)

PR #242, opened 2022-09-14
Fix for thread profiler out of bounds write

JCash:


lib/Remotery.c
```diff
@@ -5508,6 +5508,12 @@ static rmtError ThreadProfilers_GetThreadProfiler(ThreadProfilers* thread_profil
         }
     }
 
+    if (thread_profilers->nbThreadProfilers <  thread_profilers->maxNbThreadProfilers)
+    {
+        mtxUnlock(&thread_profilers->threadProfilerMutex);
+        return RMT_ERROR_MALLOC_FAIL;
+    }
```
JCash, line 5515: A project had thread leaks, which resulted in memory overwrites here.

As a repro case, use thread.c and increase number of threads to 300.
Then run with `clang -std=c11 -DRMT_ENABLED=1 -g -O0 -fsanitize=undefined -Ilib sample/thread.c lib/Remotery.c -o build/test && ./build/test 2000` to get this info from UBSan:
```
lib/Remotery.c:5470:77: runtime error: index 257 out of bounds for type 'ThreadProfiler [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior lib/Remotery.c:5470:77 in 
```

Thanks to @AGulev for the finding of this issue.

lib/Remotery.c
```diff
@@ -5508,6 +5508,12 @@ static rmtError ThreadProfilers_GetThreadProfiler(ThreadProfilers* thread_profil
         }
     }
 
+    if (thread_profilers->nbThreadProfilers+1 > thread_profilers->maxNbThreadProfilers)
+    {
+        mtxUnlock(&thread_profilers->threadProfilerMutex);
+        return RMT_ERROR_MALLOC_FAIL;
+    }
```
JCash, line 5515: A project had thread leaks, which resulted in memory overwrites here.

As a repro case, use thread.c and increase number of threads to 300.
Then run with clang -std=c11 -DRMT_ENABLED=1 -g -O0 -fsanitize=undefined -Ilib sample/thread.c lib/Remotery.c -o build/test && ./build/test 2000 to get this info from UBSan:

lib/Remotery.c:5470:77: runtime error: index 257 out of bounds for type 'ThreadProfiler [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior lib/Remotery.c:5470:77 in 

Thanks to @AGulev for the finding of this issue.

JCash: Yes, they accidentally created very many threads.

dwilliamson: Oh, ok... though it occurs to me that any process recreating many threads throughout its lifetime will no longer be able to profile.

PR #248, opened 2023-07-10
Support a single client connecting to multiple servers/processes

### LINKED ISSUE #134: Support a single client connecting to multiple servers/processes
Should be fairly straight-forward, requiring two steps:

1. Server increments base port number on failure to `bind`.
2. Client continuously polls a range of ports for connection.
3. Create a time sync between all connections.

ChipmunkV:
Uses address patterns like `ws:/127.0.0.1:1781{5,6}/rmt` to connect to multiple servers. Shows thread names with the `0_`, `1_`... prefixes depending on the server of origin. Sends console input to the first server.

Added GSST and SSST messages for retrieving the starting timepoints from the servers.

Copy-pasted the https://github.com/juliangruber/balanced-match and https://github.com/juliangruber/brace-expansion to get the address pattern parsing.

Not sure if it works for the Processor Timelines.

Fixes #134


dwilliamson:
Thank you for this PR. I will have a close look at its implications.

Not to worry about processor timeline view, as long as it doesn't break.

dwilliamson:
I'm trying to test the processor view right now but it doesn't actually appear to be working on the laptop I'm testing on, so I may have to debug that first before I get any further with testing this.

dwilliamson (changes requested):
With these changes, the processor view has lost its colours. I haven't verified if the processor view values are actually accurate.

image

image

PR #255, opened 2024-01-07
Fixed return value in `_rmt_BeginMetalSample` declaration

fiserj:
It seems that the [declaration](https://github.com/Celtoys/Remotery/blob/616fdf222a6b98f09de5e7fab167099dd92bef39/lib/Remotery.h#L1062) and [definition](https://github.com/Celtoys/Remotery/blob/616fdf222a6b98f09de5e7fab167099dd92bef39/lib/Remotery.c#L9757) of `_rmt_BeginMetalSample` are incompatible and fail to compile (when `RMT_USE_METAL` is on).

### DIFF
--- a/lib/Remotery.h
+++ b/lib/Remotery.h
@@ -1059,7 +1059,7 @@ RMT_API void _rmt_EndOpenGLSample(void);
 #endif
 
 #if RMT_USE_METAL
-RMT_API void _rmt_BeginMetalSample(rmtPStr name, rmtU32* hash_cache);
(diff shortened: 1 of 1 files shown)


fiserj:
Additionally, I was wondering why this wasn't caught by the CI, but it looks like the code is never built with `RMT_USE_METAL` on, so I could maybe add that into this or another PR as well (?).

dwilliamson:
Great catch! More than happy to accept any PRs for Metal tests, if you have the time.

PR #256, opened 2024-01-08
WIP: Vulkan sampling support

Valakor:
_NOTE: This pull request includes a handful of fixes that aren't strictly related to Vulkan sampling which I'll list below. They could be pulled out into a separate change if required._
1. _Fix `Remotery_Destructor`: Move clearing of `g_Remotery` to after graphics API-specific deletion/unbind calls, avoiding crashes because of unchecked access to g_Remotery._
2. _Fix `RMT_ARCH_64BIT`: Check `__arm64__` as well._
3. _Fix `rmt_GetLastErrorMessage` declaration: Enclose inside `extern "C"` block for C++ users._
4. _Fix `rmt_ScopedD3D12Sample`: Remove extra parens in end-of-scope object declaration which confuse some compilers that think it's a function declaration (most vexing parse)._
5. _Fix `Remotery.js` deserialization of `Uint64` and `Sint64` properties: `Remotery.c` already converts these to `rmtF64` prior to sending them over the socket._
6. _Add support for thread names on MacOS via `pthread_setname_np` and `pthread_getname_np`._

This is a first-pass implementation of Vulkan sampling support for Remotery based on the D3D12 implementation. It currently expects a handful of things about the user's Vulkan environment that are likely more restrictive than necessary but make the implementation much simpler.

- The user must either use Vulkan 1.2 with the `hostQueryReset` and `timelineSemaphore` features enabled, or have enabled the extensions `VK_EXT_host_query_reset` and `VK_KHR_timeline_semaphore`. These features/extensions are pretty widely supported so it seems maybe reasonable to require that users enable them; without them the implementation gets reasonably more complex and less performant.
- The user must enable the extension `VK_EXT_calibrated_timestamps`. It'd probably be pretty easy to make usage of this extension optional.
- The implementation always loads Vulkan functions dynamically. This is generally the most-preferred method, but it might be nicer to instead just have the user provide all the functions as a configuration option up-front. Unfortunately that's difficult without a bunch of horrible forward declarations, or including Vulkan.h in Remotery.h.
- The way I'm passing `vkGetInstanceProcAddr` feels very hacky, definitely open to suggestions here.
- I've left a handful of `TODO(valakor)`'s in the code for potential future improvement points.
- I've tried to do everything to support Windows+MacOS+Linux but it's possible I missed some things on non-Windows platforms. Mac is especially tricky since the normal Vulkan API doesn't understand that Metal doesn't support writing timestamps inside render/compute passes, but this is true for any user of Vulkan via MoltenVK.

I'm open to discussion/feedback about this implementation!

### DIFF
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -189,6 +189,10 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
     #include <atomic>
 #endif
 
+#ifndef RMT_PLATFORM_WINDOWS
+    #define ZeroMemory(dest, len) memset(dest, 0, len)
+#endif
+
 // clang-format on
 
 #if defined(_MSC_VER) && !defined(__clang__)
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -7098,12 +7098,6 @@ static void Remotery_Destructor(Remotery* rmt)
 
     rmtDelete(ThreadProfilers, rmt->threadProfilers);
 
-    if (g_RemoteryCreated)
-    {
-        g_Remotery = NULL;
-        g_RemoteryCreated = RMT_FALSE;
-    }
-
     rmtDelete(ObjectAllocator, rmt->propertyAllocator);
 
 #if RMT_USE_VULKAN
@@ -7129,6 +7123,12 @@ static void Remotery_Destructor(Remotery* rmt)
 #if RMT_USE_METAL
     rmtDelete(Metal, rmt->metal);
 #endif
+
+    if (g_RemoteryCreated)
+    {
+        g_Remotery = NULL;
+        g_RemoteryCreated = RMT_FALSE;
+    }
 
     rmtCloseFile(rmt->logfile);
 
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -10441,7 +10441,7 @@ static rmtError VulkanMarkFrame(VulkanBindImpl* bind)
     return RMT_ERROR_NONE;
 }
 
-RMT_API rmtError _rmt_BindVulkan(void* instance, void* physical_device, void* device, void* queue, VulkanGetInstanceProcAddr get_instance_proc_addr, rmtVulkanBind** out_bind)
+RMT_API rmtError _rmt_BindVulkan(void* instance, void* physical_device, void* device, void* queue, rmtVulkanGetInstanceProcAddr get_instance_proc_addr, rmtVulkanBind** out_bind)
 {
     VulkanBindImpl* bind;
     VkInstance vulkan_instance = (VkInstance)instance;
--- a/lib/Remotery.c
+++ b/lib/Remotery.c
@@ -51,6 +51,7 @@
     @D3D12:         Direct3D 12 event sampling
     @OPENGL:        OpenGL event sampling
     @METAL:         Metal event sampling
+    @VULKAN:        Vulkan event sampling
     @SAMPLEAPI:     Sample API for user callbacks
     @PROPERTYAPI:   Property API for user callbacks
     @PROPERTIES:    Property API
@@ -392,6 +393,14 @@ static void usTimer_Init(usTimer* timer)
 #endif
 }
 
+#if defined(RMT_PLATFORM_WINDOWS)
+    #define usTimer_FromRawTicks(timer, ticks) (rmtU64)(((ticks) - (timer)->counter_start.QuadPart) * (timer)->counter_scale)
+#elif defined(RMT_PLATFORM_MACOS)
+    #define usTimer_FromRawTicks(timer, ticks) (rmtU64)(((ticks) - (timer)->counter_start) * (timer)->counter_scale)
+#elif defined(RMT_PLATFORM_LINUX)
+    #define usTimer_FromRawTicks(timer, ticks) (rmtU64)((ticks) - (timer)->counter_start)
+#endif
+
 static rmtU64 usTimer_Get(usTimer* timer)
 {
 #if defined(RMT_PLATFORM_WINDOWS)
@@ -401,18 +410,19 @@ static rmtU64 usTimer_Get(usTimer* timer)
 
     // Read counter and convert to microseconds
     QueryPerformanceCounter(&performance_count);
-    return (rmtU64)((performance_count.QuadPart - timer->counter_start.QuadPart) * timer->counter_scale);
+    return usTimer_FromRawTicks(timer, performance_count.QuadPart);
 
 #elif defined(RMT_PLATFORM_MACOS)
 
     rmtU64 curr_time = mach_absolute_time();
-    return (rmtU64)((curr_time - timer->counter_start) * timer->counter_scale);
+    return usTimer_FromRawTicks(timer, curr_time);
 
 #elif defined(RMT_PLATFORM_LINUX)
 
     struct timespec tv;
     clock_gettime(CLOCK_REALTIME, &tv);
-    return ((rmtU64)(tv.tv_sec * (rmtU64)1000000) + (rmtU64)(tv.tv_nsec * 0.001)) - timer->counter_start;
+    rmtU64 ticks = (rmtU64)(tv.tv_sec * (rmtU64)1000000) + (rmtU64)(tv.tv_nsec * 0.001);
+    return usTimer_FromRawTicks(timer, ticks);
 
 #endif
 }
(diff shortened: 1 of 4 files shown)


lib/Remotery.c
```diff
@@ -10091,7 +10091,8 @@ typedef struct VulkanBindImpl
     VulkanSample** sampleRingBuffer;
 
     // Read/write positions of the ring buffer allocator, synchronising access to all the ring buffers at once
-    // TODO(don): Separate by cache line?
+    // NOTE(valakor): These are 64-bit instead of 32-bit so that we can reasonably assume they never wrap.
```
dwilliamson, line 10108: Why is this and is there any code that will break if these positions wrap?

(is it the reference to ABA?)

lib/Remotery.c
```diff
@@ -189,6 +189,10 @@ static rmtBool g_SettingsInitialized = RMT_FALSE;
     #include <atomic>
 #endif
 
+#ifndef RMT_PLATFORM_WINDOWS
+    #define ZeroMemory(dest, len) memset(dest, 0, len)
```
dwilliamson, line 193: Can you just use `memset` in Vulkan code instead of overloadnig this symol?

lib/Remotery.c
```diff
@@ -10280,18 +10280,26 @@ static rmtError GetTimestampCalibration(VulkanBindImpl* bind, VkPhysicalDevice v
...
+	// TODO(valakor): We have to fall back to manually querying CPU time due to the following issue:
+    //  On Apple platforms MoltenVK reports support for VK_TIME_DOMAIN_CLOCK_MONOTONIC_RAW_EXT, which matches the time
+	//  domain of mach_continuous_time(). To support mach_absolute_time() Vulkan would have to extend the available
+	//  time domains to include something like "VK_TIME_DOMAIN_CLOCK_UPTIME_RAW_EXT". See the comments here:
+	//  https://github.com/KhronosGroup/MoltenVK/blob/main/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+	//
+	//  Alternatively, Remotery could switch to using mach_continuous_time(). The difference between the two is that
+    //  mach_continuous_time() (CLOCK_MONOTONIC_RAW) includes system sleep time, whereas mach_absolute_time()
+	//  (CLOCK_UPTIME_RAW) does not. I'm not 100% convinced that's what we would want, but I think it is technically
+	//  more secure.
```
dwilliamson, line 10302: Remotery should definitely be switched to using `mach_continuous_time` if it makes all this easier.

dwilliamson (commented):
This all looks great. Not really much to critique, though I did go through each commit and try to use my very limited Vulkan knowledge to ensure all was OK.

lib/Remotery.c
Valakor: Definitely - I was just being lazy...

lib/Remotery.c
Valakor: If these counters wrap then a whole bunch of existing code that does comparisons like `current_write_gpu > current_read_cpu` breaks. It could be rewritten with some amount of effort (maybe), but using 64bit counters avoids the problem in the first place (for any reasonable program).

For example, a program running at 1000fps that writes 200 timestamps per frame will wrap a 32-bit counter in ~6 hours, whereas a 64-bit counter won't wrap for almost 3 million years.

The D3D12 code still uses 32-bit counters but could be modified without too much effort if we wanted to.

lib/Remotery.c
Valakor: I don't know the right answer here honestly. Things are extra complicated because the internal function provided by Metal that MoltenVK is wrapping here actually returns values in the mach_absolute_time() time domain which would be preferable, but this can't be returned via the Vulkan function because there's no equivalent Vulkan time domain (CLOCK_UPTIME_RAW_EXT)...

Valakor: Regardless of what we do on MacOS I think Remotery should probably change what clock it uses on Linux to `CLOCK_MONOTONIC_RAW` though that would complicate Vulkan timestamp calibration for Linux since I don't think it's necessarily guaranteed to be available...

All of this makes me want to drop all this special-casing and just query our CPU timer manually right after getting a device timestamp, which is what UE5 seems to do... I honestly don't know enough about the behavior of timestamp calibration across enough devices to know what the most reasonable approach is.

lib/Remotery.c
Valakor: Also, Vulkan timeline semaphores are 64-bit counters that must strictly increase so they're basically not allowed to wrap.

lib/Remotery.c
dwilliamson: Ideally we would use `CLOCK_MONOTONIC_RAW` but I am not versed enough in how widely supported that is. Remotery works as far as OpenBSD and I'm keen to keep it that way.

lib/Remotery.c
Valakor: The culprit is `VulkanMarkFrame` (and `D3D12MarkFrame`) where there are comparisons between `current_write_gpu`, `current_read_cpu`, and `msg_sample_tree->userData`.

That aside, Vulkan still requires strictly increasing values for semaphores. I guess this could be avoided by creating a new semaphore when things are about to wrap but that feels more complicated than just using 64-bit counters.

dwilliamson: OK, that's a bug! I will either fix that or move the D3D12 ones to 64-bit indices.

dwilliamson:
What's left to pull this out of Draft?

Valakor:
> What's left to pull this out of Draft?

I think it might be worth discussing the API of `rmt_BindVulkan` and whether the user should provide function pointers to Remotery or whether Remotery should continue to look them up dynamically. I'm not 100% sure if the current implementation will work in cases where the user links to Vulkan statically; having the user provide the function pointers would allow Remotery to support a wider range of Vulkan implementations / environments. Unfortunately, doing this would probably require including Vulkan.h in the Remotery header to avoid super gross forward declarations which I wasn't sure was acceptable - it's a pretty big include and `Remotery.h` is currently fairly lightweight.

It's probably also worth discussing whether it's worth relaxing the current Vulkan version / extension requirements. If you're fine with the current limitations then it could probably be merged in its current form and future changes could be taken to support older Vulkan implementations.

dwilliamson:
> I think it might be worth discussing the API of `rmt_BindVulkan` and whether the user should provide function pointers to Remotery or whether Remotery should continue to look them up dynamically. I'm not 100% sure if the current implementation will work in cases where the user links to Vulkan statically; having the user provide the function pointers would allow Remotery to support a wider range of Vulkan implementations / environments. Unfortunately, doing this would probably require including Vulkan.h in the Remotery header to avoid super gross forward declarations which I wasn't sure was acceptable - it's a pretty big include and `Remotery.h` is currently fairly lightweight.
> 
> _Edit: Somehow this whole time I missed that Remotery already has an answer to this problem for CUDA - declaring all the required function pointers as `void*` passed via a struct. It would be fairly simple for me to change the Vulkan code to something similar. I could also model the Vulkan error handling similarly to CUDA if requested by mapping Vulkan error codes to new `rmtError` codes._
> 

By static linking, I presume you mean platforms like iOS? This issue here seems to imply you can link statically and use `vkGetInstanceProcAddr` as well: https://github.com/KhronosGroup/MoltenVK/issues/1809

Also, do you have to pass the `vkGetInstanceProcAddr` function to `rmt_BindVulkan`? We manage to avoid this with OpenGL and `wglGetProcAddress`, but only really because each platform has its own OpenGL function getter.

> It's probably also worth discussing whether it's worth relaxing the current Vulkan version / extension requirements. If you're fine with the current limitations then it could probably be merged in its current form and future changes could be taken to support older Vulkan implementations.

I think it's not worth worrying about. If anybody needs extra functionality and you personally do not need it right now, it's worth waiting until somebody adds a PR for it.

dwilliamson:
That's pretty much what we do with OpenGL.

However, in the absence of a wide array of platforms to test this on, currently, I'm more than happy to accept a PR that requires you to pass Vulkan function pointers as part of `rmt_BindVulkan`.

Valakor:
I think I'll submit changes to have the user pass the function pointers. I realized when I was doing some final cleanup that many (all?) Vulkan implementations might expose function pointers for extensions that aren't enabled (or not even supported) by the implementation, and there's no way to check for extension enable state.

That might be fine if we just expect the user to have enabled the right extensions ahead of time, but unfortunately there's no way to tell whether they're using an extension or promoted feature, e.g. an EXT extension promoted to KHR or any extension promoted to core functionality in a later Vulkan version.

dwilliamson:
The new fn registration looks good, we can make smoother API changes later if required, but I doubt we'll need it.

The only build failures are typical Github spurious fails that have nothing to do with the code.

This is some great work; thanks for pushing it through!

PR #266, opened 2025-07-09
CI toolset bump

ntfshard:
Seems versions in github runner are outdated, I put default settings to make it alive

I'm not a devops person, so I potentially can do something in a not intendent by initial idea way
Android build fails on a code line, not my field of expertise, sorry

### DIFF
--- a/.github/scripts/compile.bat
+++ b/.github/scripts/compile.bat
@@ -1,5 +1,3 @@
-echo off
-
 set PLATFORM=%1
 set ARCH=%2
 set BUILDMODE=%3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up JDK 17
+      - name: Set up JDK
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin' # See 'Supported distributions' for available options
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
         arch: ['x86', 'x64']
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -47,14 +46,12 @@ jobs:
...
         buildmode: ['debug', 'release']
-        toolset: ['14.3','14.2','14.1']
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - uses: ilammy/msvc-dev-cmd@v1
         with:
           arch: ${{ matrix.arch }}
-          toolset: ${{ matrix.toolset }}
```
ntfshard, line 57: https://github.com/ilammy/msvc-dev-cmd?tab=readme-ov-file#inputs

ntfshard:
Hm.. It fells like old file was started. 
I made PR to my fork with same changes: https://github.com/ntfshard/Remotery/pull/3 (https://github.com/ntfshard/Remotery/actions/runs/16171921319?pr=3)

.github/scripts/compile.bat
```diff
@@ -1,5 +1,3 @@
-echo off
```
ntfshard, line 1: to shut down output it should be `@echo off`
imho explicit is better

.github/workflows/build.yml
```diff
@@ -71,17 +68,17 @@ jobs:
         ndkversion: [21]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
-      - name: Set up JDK 1.11
+      - name: Set up JDK 17
```
skorobogatydmitry:
```suggestion
      - name: Set up JDK
```

Why bother? I don't think it's worth duplicating.

ntfshard: Kept original style, but maybe w/o version is cleaner. Thanks

